PR #4, opened 2024-09-10
Feature: ynBNB mainnet

xhad:
This adds a deployment of the vault factory to BSC mainnet with some minor updates to constants, contract addresses, and actors with extra basic tests.

NOTES
This has two contracts, Factory and SingleVault. Vaults are deployed via the factory, and require `1 ether` of asset sent to the factory, which is used to bootstrap deposits into the newly created vault.

`SingleVault` has  TimelockController Admin capabilities that are Handled via the `YnBscSecurityCouncil` multisig:
https://app.safe.global/home?safe=bnb:0x721688652DEa9Cabec70BD99411EAEAB9485d436

ROADMAP
1. Deploy `ynBNB` as a SingleVault with `slisBNB` as the underlying asset. Full deposits and withdraws are capable, but we must not deposit into the Karak Vault until we update Karak withdraws.
2. Create KslisBNB Strategy Vault
3. Update MetaVault to deposit into KslisBNB Strategy Vault


src/SingleVault.sol
```diff
@@ -3,29 +3,22 @@ pragma solidity ^0.8.24;
...
     ReentrancyGuardUpgradeable,
     TimelockControllerUpgradeable,
-    IERC20
+    IERC20,
+    IERC4626,
+    Math
 } from "src/Common.sol";
 
 import {ISingleVault} from "src/ISingleVault.sol";
+import {SlisBnbModule} from "src/module/SlisBnbModule.sol";
```
danoctavian, line 14: This should be functionally equivalent to having the code in contract as before.

The library is deployed external so it doesn't increase the bytecode of the contract, and in theory can be called by other contracts

But must be noted to upgrade the logic in the module we still need to upgrade the whole implementation of the SingleVault.

Also because of how libraries work in solidity, there's no way to have a library interface so it makes this contract import the concrete library so SingleVault can't be generic (as opposed to it depending on a generic IBnbModule contract)

src/module/SlisBnbModule.sol
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * @dev Retrieves the ERC4626Storage struct from a given storage location.
+     * from oz/ERC4626Upgradeable.sol because its private
+     * @return $ The ERC4626Storage struct stored at the given location.
+     */
+    function _retrieveERC4626Storage() internal pure returns (ERC4626Storage storage $) {
+        assembly {
+            $.slot := 0x0773e532dfede91f04b12a73d3d2acd361424f41f76b4fb79f090161e36b4e00
```
danoctavian, line 33: nice find here - this is actually the only way to access storage inside a library without passing a storage reference

src/SingleVault.sol
kahuna099: Yeah, deploying a new SingleVault which has an updated SlisBnbModule will require redeploying the updated SlisBnbModule. But this should all be handled by foundry once you deploy the new SingleVault implementation (which imports the new SlisBnbModule), after which you upgrade the proxy to point to this new impl.

xhad: Yes, agree. Thanks for this review. I've reverted efforts to make the SingleVault generic and will just deploy the basic functionality needed for ynBNB now.

src/module/SlisBnbModule.sol
xhad: I like how Libraries are pure and consume storage from specified slots like this. I'll come back to this research after deploying the base version. Again, I've reverted back to the basic implementation we had previously and will focus on how to generalize later.

danoctavian: Makes sense to keep it minimal for now but good that you explored this direction

PR #6, opened 2024-10-18
Feature: Max Vault V1

xhad:
This is a super rough early draft of the outline and code structure for review. 

The deposit function should be working with `forge test` using WETH as the base underlying asset. 

1. Uses it's own rate provider to fetch asset prices, in the modules folder
2. Uses etches to mock the WETH contract on mainnet 
3. Makes use of the Storage library to offload the view and state fetching

Still working on the Accounting for this, but want to get your eyes on it early


src/Vault.sol
```diff
@@ -0,0 +1,261 @@
...
+    IERC20,
+    Math,
+    ReentrancyGuardUpgradeable,
+    SafeERC20
+} from "./Common.sol";
+
+import {Storage} from "src/Storage.sol";
+import {IVault} from "src/interface/IVault.sol";
+
+contract Vault is IVault, ERC20PermitUpgradeable, AccessControlUpgradeable, ReentrancyGuardUpgradeable {
```
danoctavian, line 17: what's the intention on how to handle withdrawals for the Vault?

danoctavian, line 17: What's the intention on how to handle allocation execution?

Example:

100 ETH is deposited.

Vault

deposits  10 ETH to ynETH
converts 50 ETH to stETH and allocates it to ynLSDE
deposits 40 ETH to A Third party ERC4626 strategy

danoctavian, line 17: What's the intention on how to handle 

ynETH and ynLSDe existing yieldnest strategies?

Can they simply be classifed as deposit assets for the vault so that current existing ynETH and ynLSDe be directly deposited to ynETHx?

xhad: async: A simple withdraw queue would likely be the best phase 1 implementation
sync: Most projects use a buffer of underlying (WETH) that sits in the vault

danoctavian: Suppose that there are 2 strategies that can take stETH

ynLSDe (eigenlayer)
ynLSDk (karak)

How can the vault decide to distribute assets here ?

src/Vault.sol
```diff
@@ -0,0 +1,261 @@
...
+
+        // QUESTION: Keep the balances for the assets, or keep that balances in base price
+
+        // NOTE: Get the loops out of the public calls
+
+        //yv3
+        // what if you want to udpate a single strategy, but what if you want to update one strat
+        // the debt value is used to separate the deposited value, you know what the rewards
+        // balance is based on the the debt less the the rewards.
+        // allows you to not have to grind through everything??
```
danoctavian, line 255: Other things to consider here:

Suppose we have strategies with asynchronous withdrawals:

ynETH
https://etherscan.io/address/0x0BC9BC81aD379810B36AD5cC95387112990AA67b

Example: weETH (hypothetical)
https://etherscan.io/address/0x7d5706f6ef3F89B3951E23e557CDFBC3239D4E2c

What would be an approach to handle totalAssets assuming one of these positions is being unwinded?

Since across withdrawal mechanisms a standard does not exist.

Some may not use NFTs and simply positions so the behaviour is custom.

How. can we code that behaviour to still measure totalAssets?

src/Vault.sol
danoctavian: Also each strategy has potentially different deposit signature calls

Eg.

deposit(address. asset, uint256 amount, address receiver)
deposit(uint256 amount, address receiver)
or some other version.

In the beginning the `bytes calldata` parameter was considered to inject custom calls with some guard rails around those.

We'd have to have a way of doing that

src/Vault.sol
```diff
@@ -0,0 +1,261 @@
...
+    function deposit(uint256 assets_, address receiver) public returns (uint256) {
+        if (paused()) revert Paused();
+
+        uint256 shares = previewDeposit(assets_);
+        _deposit(asset(), _msgSender(), receiver, assets_, shares);
+
+        return shares;
+    }
+
+    function depositAsset(address asset_, uint256 assets_, address receiver) public returns (uint256) {
```
danoctavian, line 108: we may consider overloading here with 
deposit(address asset_, uint256 assets_, address receiver)

src/Vault.sol
danoctavian: As discussed in the call, sounds like we'd want

the strategy controller/governor  to be able to call a limited set of transactions  with whitelisted targets

src/Vault.sol
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    // QUESTION: What params should be used here? strategy, asset, etc.
+    function processAccounting() public {
+        // get the balances of the assets
+        // AssetStorage storage assetStorage = _getAssetStorage();
+
+        // for (uint256 i = 0; i < assetsStorage.list.length; i++) {
+        //     address asset = assets[i];
+        //     idleBalance = asset.balanceOf(address(vault));
```
danoctavian, line 241: balanceOf is the way to go to easily handle stETH and oETH (rebasing)

src/Vault.sol
```diff
@@ -0,0 +1,261 @@
...
+        // }
+        // get the balances of the strategies
+
+        // call convertToAssets on the strategie?
```
danoctavian, line 245: assuming the strategies are ERC20s this can actually simplify the logic here 

and push the logic of calculating the value of X amount of Shares of Strategy to the rate provider

danoctavian: Another case here is if you're not dealing with a strategy that's an ERC20.

It can be an NFT (eg. a Uniswap V3 position)

A pending withdrawal is similar to that as well.

so potentially we'd have to have a TVL module here to customize for things that are not ERC20s

src/Vault.sol
```diff
@@ -0,0 +1,261 @@
+// SPDX-License-Identifier: BSD Clause-3
+pragma solidity ^0.8.24;
```
danoctavian, line 2: overall approach  looks good 
 
 1.starting off with this mixed ERC4626 and multi-asset vault interface
 2. build the rest of the logic
 
 figure out later what separate modules we need to add based on what is common and what's different or looks like it needs to be upgraded

src/interface/IVault.sol
```diff
@@ -0,0 +1,88 @@
...
+        // QUESTION: Why was the enzyme 
+        address rateProvider;
+        bool paused;
+    }
+
+    struct AssetParams {
+        bool active;
+        uint256 index;
+        uint8 decimals;
+        uint256 idleAssets;
```
danoctavian, line 32: do we still need idleAssets and deployedAssets when we use asset.balanceOf(address(vault)) to measure what's in the vault?

Also storing idleAssets/deployed assets won't work for rebasing tokens stETH/oETH

src/interface/IVault.sol
```diff
@@ -0,0 +1,88 @@
...
+
+    struct AssetStorage {
+        mapping(address => AssetParams) assets;
+        address[] list;
+    }
+
+    struct StrategyParams {
+        bool active;
+        uint256 index;
+        uint256 ratio;
```
danoctavian, line 44: if we allow the governor/strategy controller to do arbitrary strategy allocations

do we still need the ratio ?

src/Vault.sol
xhad: > Suppose that there are 2 strategies that can take stETH
> 
> ynLSDe (eigenlayer) ynLSDk (karak)
> 
> How can the vault decide to distribute stETH assets here ?
> 
> Eg: 100 stETH enters the vault, and those 2 stETH strategies exist

Perhaps we can start with a permissioned Governor function that is able to deposit or withdraw from strategy vaults.

xhad: > Also each strategy has potentially different deposit signature calls
> 
> Eg.
> 
> deposit(address. asset, uint256 amount, address receiver) deposit(uint256 amount, address receiver) or some other version.
> 
> In the beginning the `bytes calldata` parameter was considered to inject custom calls with some guard rails around those.
> 
> We'd have to have a way of doing that

Yeah, I guess we can whitelist to only call strategy contracts as targets, and will have to experiment with how to make the func signatures safe, but generalized.

xhad: > As discussed in the call, sounds like we'd want
> 
> the strategy controller/governor to be able to call a limited set of transactions with whitelisted targets

How about:
Governor sets the strategies:
operator can call any function on the strategies with calldata, but txes are restricted to only calling the strategy contracts.

xhad: > Also each strategy has potentially different deposit signature calls
> 
> Eg.
> 
> deposit(address. asset, uint256 amount, address receiver) deposit(uint256 amount, address receiver) or some other version.
> 
> In the beginning the `bytes calldata` parameter was considered to inject custom calls with some guard rails around those.
> 
> We'd have to have a way of doing that

yes, I'll take a look at how to implement this and ping you with it today or tomorrow

src/Vault.sol
```diff
@@ -0,0 +1,398 @@
...
+
+    function _getStrategyStorage() internal pure returns (StrategyStorage storage $) {
+        assembly {
+            $.slot := 0x36e313fea70c5f83d23dd12fc41865566e392cbac4c21baf7972d39f7af1774d
+        }
+    }
+
+    //// ADMIN ////
+
+    function setRateProvider(address rateProvider_) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
danoctavian, line 322: can be done at the end but i'd keep DEFAULT_ADMIN_ROLE only as a role manager,

and have decoupled roles for these set calls, to allow flexibiliy

src/Vault.sol
```diff
@@ -0,0 +1,398 @@
...
+    }
+
+    function maxWithdraw(address owner) public view returns (uint256) {
+        if (paused()) return 0;
+
+        // Here we return whichever is less, the available base assets in the buffer, or the user's
+        // vault token balance converted to base asset value
+        uint256 ownerAssets = _convertToAssets(asset(), balanceOf(owner), Math.Rounding.Floor);
+        // The buffer strategy must have the same base underlying asset as the vault
+        uint256 idleBalance = IERC20(asset()).balanceOf(address(this));
```
danoctavian, line 62: questioning here if we should use the freshly deposited funds to fulfil withdrawals.

It's more controlled if withdrawers only have access to the buffer funds, not to fresh deposits as well.

CC @gearcat0  do you have more opinions here on withdrawing funds that were just deposited

src/Vault.sol
```diff
@@ -0,0 +1,398 @@
...
+        return _getStrategyStorage().strategies[asset_];
+    }
+
+    function previewDepositAsset(address asset_, uint256 assets_) public view returns (uint256) {
+        if (!_getAssetStorage().assets[asset_].active) revert InvalidAsset();
+        return _convertToShares(asset_, assets_, Math.Rounding.Floor);
+    }
+
+    function depositAsset(address asset_, uint256 assets_, address receiver) public nonReentrant returns (uint256) {
+        if (paused()) revert Paused();
```
danoctavian, line 164: personal preference for 

if (paused()) {
revert Paused();
}

as it allows no room for error when adding a second line

src/Vault.sol
```diff
@@ -0,0 +1,398 @@
...
+            StrategyParams({active: true, index: index, decimals: decimals_, idleBalance: 0});
+
+        strategyStorage.list.push(strategy);
+
+        emit NewStrategy(strategy, index);
+    }
+
+    function pause(bool paused_) external onlyRole(DEFAULT_ADMIN_ROLE) {
+        VaultStorage storage vaultStorage = _getVaultStorage();
+        if (_getVaultStorage().rateProvider == address(0)) revert RateProviderNotSet();
```
danoctavian, line 377: the conditionals around pausing are a bit confusing.

it is correct if it starts paused, but in general you'd expect pause/unpause does just that and nothing else

src/Vault.sol
```diff
@@ -0,0 +1,398 @@
...
+            assetStorage.assets[asset_].idleBalance = idleBalance;
+            totalBaseBalance += _convertAssetToBase(asset_, idleBalance);
+        }
+
+        for (uint256 i = 0; i < strategyStorage.list.length; i++) {
+            address strategy = strategyStorage.list[i];
+            uint256 idleBalance = IERC20(strategy).balanceOf(address(this));
+            strategyStorage.strategies[strategy].idleBalance = idleBalance;
+            totalBaseBalance += _convertAssetToBase(strategy, idleBalance);
+        }
```
danoctavian, line 202: we have to handle the asynchronous withdrawal positions for certain strategies with custom logic for ynETH and ynLSDe and potentially other asynchronous withdrawals strategies.

For ynETH and ynLSDe, you'd call 

withdrawalRequestsForOwner
https://github.com/yieldnest/yieldnest-protocol/blob/main/src/WithdrawalQueueManager.sol#L561
and sum the WithdrawalRequest.amount

For Karak you have Vault.getQueuedWithdrawal (no NFTs)
https://github.com/karak-network/v2-contracts/blob/main/src/Vault.sol#L258


In contrast, Kernel will be deployed with instant withdrawals.



https://github.com/Kelp-DAO/kernel-smart-contracts-public/blob/main/src/KernelVault.sol#L132

Going into this, there was the idea of packaging karak in an ynLSDk asset on its own. That i think may still makes sense since you have to not only handle staking/unstaking but also operator selection. With karak you create one vault per validator(?).

Regardless, It seems like here we'd need the custom module/modules functionality to handle withdrawal positions. In theory this way you could plug in other protocols as well as strategies (Eg. you can plug in WEETH as a tokenized strategy). 

For these modules we can either use inheritance (inherit this vault and just add the logic for each strategy) or additional Module contracts

src/Vault.sol
```diff
@@ -0,0 +1,398 @@
...
+
+    function setBufferStrategy(address bufferStrategy_) external onlyRole(DEFAULT_ADMIN_ROLE) {
+        if (bufferStrategy_ == address(0)) revert ZeroAddress();
+        StrategyStorage storage strategyStorage = _getStrategyStorage();
+        if (!strategyStorage.strategies[bufferStrategy_].active) revert InvalidStrategy();
+        _getVaultStorage().bufferStrategy = bufferStrategy_;
+        emit SetBufferStrategy(bufferStrategy_);
+    }
+
+    function addAsset(address asset_, uint8 decimals_) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
danoctavian, line 336: we can consider reading asset.decimals() directly from asset, grand majority support that function

src/Vault.sol
```diff
@@ -0,0 +1,398 @@
...
+    function processAllocation(address[] calldata targets, uint256[] memory values, bytes[] calldata data)
+        external
+        onlyRole(DEFAULT_ADMIN_ROLE)
+    {
+        for (uint256 i = 0; i < targets.length; i++) {
+            bool success;
+            bytes memory returnData;
+
+            // Determine the type of target and process accordingly
+            if (_getAssetStorage().assets[targets[i]].active) {
```
danoctavian, line 216: the active boolean is used to guard both deposit calls and processAllocation.

We may want to turn off an asset but still manage it with processAllocation.

Not sure it's needed here at all,  and if the flag is needed it should be decoupled from the one guarding deposits

src/Vault.sol
```diff
@@ -0,0 +1,398 @@
...
+                // Process asset approvals // QUESTION: Assumiung we only need approvals for assets
+                if (data[i].length > 0 && data[i][0] == abi.encodeWithSignature("approve(address,uint256)")[0]) {
+                    (success, returnData) = targets[i].call{value: values[i]}(data[i]);
+                    if (!success) revert ProcessFailed(returnData);
+                }
+            } else if (_getStrategyStorage().strategies[targets[i]].active) {
```
danoctavian, line 222: this doesn't work for handling wihdrawals on ynETH/ynLSDe currently as you interact with a different contract.

We can potentially consider to upgrade those to ebmed withdrawals into the token assets, but even then not all strategies support calls targeted at a single contract (eg. etherfi WEETH)

src/Vault.sol
```diff
@@ -0,0 +1,398 @@
...
+                (success, returnData) = targets[i].call{value: values[i]}(data[i]);
+                if (!success) revert ProcessFailed(returnData);
+
+                // Update idle balance for the strategy
+                uint256 idleBalance = IERC20(targets[i]).balanceOf(address(this));
+                _getStrategyStorage().strategies[targets[i]].idleBalance = idleBalance;
+            } else {
+                // Revert if the target is neither an active asset nor an active strategy
+                revert BadStrategy(targets[i]);
+            }
```
danoctavian, line 233: in addition to strategies i can see the use case for

swapping
wrapping/unwrapping (eg. oeth to WOETH, stETH to wsteth)

 send funds to the buffer strategy
 
 I wonder if we should just have a  generic whitelist of targets that we update.
 
 Additionally, parameters matter a lot. You may whitelist uniswap, but that would still allow swaps to a random shitcoin

src/Vault.sol
gearcat0: I don't object for the vault

src/Vault.sol
xhad: I've updates the vault docs here in this repo to explain how withdraws function. Closing this for now

xhad: Picking up these questions in the scenario tests folder

xhad: These will be assets and strategies. Will pick up this question in the scenario tests for ynETH and ynLSDe

src/Vault.sol
xhad: Will pick up this Q in the ynETH scenario test

src/Vault.sol
xhad: implemented this suggestion

src/Vault.sol
xhad: the intention was to prevent unpausing unless the rateProvider and bufferStrategy have been setup

src/Vault.sol
xhad: great feedback. I've refactored how the processor makes calls by adding a ProcessorStorage position that tracks a whitelisting of function signatures on contracts:

https://github.com/yieldnest/yieldnest-vault/blob/0ec0acf6d968612a3611127f67017d9c976d2306/src/Vault.sol#L219

src/Vault.sol
xhad: my concern is with handling USDC, which I suspect we will with ynUSDx

src/Vault.sol
xhad: I've refactored this processor function, interested to get your feedback
https://github.com/yieldnest/yieldnest-vault/blob/0ec0acf6d968612a3611127f67017d9c976d2306/src/Vault.sol#L219

src/Vault.sol
xhad: yeah fixed this, see commented above

src/interface/IVault.sol
xhad: ratio has been removed. see the processor function

src/Vault.sol
xhad: yeah good feedback, I'm going to leave the roles to the end once I see how everything shapes up

src/Vault.sol
xhad: I've removed taking withdraws from idleBalances because it doubled the gas cost for withdraws. Now all withdraws come from the buffer strategy and get send to the owner from the strategy vault.

src/Vault.sol
danoctavian: valid concern

https://etherscan.io/address/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48#readProxyContract

USDC has 6 decimals. so does USDT and Paypal USD etc.

We'll have to decide what ynUSDx has - i'd go for 18 like DAI, Ethena and the Defi Stables. It would still work to have our vaults have 18 decimals and assets with different decimals so it

I noticed one gotcha here though: 

https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/723f8cab09cdae1aca9ec9cc1cfa040c2d4b06c1/contracts/token/ERC20/extensions/ERC4626Upgradeable.sol#L98C30-L98C43

OZ default ERC4626 implementation takes the decimal count from whatever you configure as the asset - we'd need to override that behaviour

src/Vault.sol
```diff
@@ -0,0 +1,430 @@
...
+            strategyStorage.strategies[strategy].idleBalance = idleBalance;
+            totalBaseBalance += _convertAssetToBase(strategy, idleBalance);
+        }
+
+        _getVaultStorage().totalAssets = totalBaseBalance;
+
+        // QUESTION: Should we have a bounty to call this function? Perhaps a withdraw fee that goes to msg.sender here
+    }
+
+    function processor(address[] calldata targets, uint256[] memory values, bytes[] calldata data)
```
danoctavian, line 220: we can also return an array of  `bytes[] memory returnData`

for each call to make return values accessible

In theory you can thus have a smart contract on top of this function that does sequences of actions and acts upon return values

xhad: implemented this

src/Vault.sol
xhad: Yes, this will need to get added to the scenario testing for a USD vault.

src/Vault.sol
```diff
@@ -247,60 +356,84 @@ contract Vault is IVault, ERC20PermitUpgradeable, AccessControlUpgradeable, Reen
...
-    //     uint256 rate = IRateProvider(rateProvider()).getRate(asset_);
-    //     return baseAmount.mulDiv(10 ** getAsset(asset_).decimals, rate, Math.Rounding.Floor);
-    // }
-
-    function _decimalsOffset() internal pure returns (uint8) {
-        return 0;
-    }
-
+        uint256 assets = shares.mulDiv(totalAssets() + 1, totalSupply() + 10 ** 0, rounding);
+        uint256 baseAssets = _convertAssetToBase(asset_, assets);
```
danoctavian: this still needs to be fixed with a function that does _covertBaseToAsset(asset_, assets)

src/Vault.sol
```diff
@@ -309,97 +442,112 @@ contract Vault is IVault, ERC20PermitUpgradeable, AccessControlUpgradeable, Reen
...
         emit Pause(paused_);
     }
 
+    /**
+     * @notice Processes the accounting of the vault by calculating the total base balance.
+     * @dev This function iterates through the list of assets, gets their balances and rates,
+     *      and updates the total assets denominated in the base asset.
+     */
+    function processAccounting() public {
+        uint256 totalBaseBalance = address(this).balance;
```
danoctavian: should we put the counting of `address(this).balance` under a configurable flag? (trackNativeAsset, countNativeAsset)

It doesn't apply to non-ETH and non-BNB vaults so could be forgotten in there when forking

src/module/Provider.sol
```diff
@@ -0,0 +1,95 @@
...
+
+    function otherAssets(address vault, address strategy) public view returns (uint256 assets) {
+        if (strategy == MC.YNETH) {
+            (uint256[] memory withdrawalIndexes, WithdrawalRequest[] memory requests) =
+                IynETHwm(MC.YNETH_WM).withdrawalRequestsForOwner(vault);
+
+            uint256 length = withdrawalIndexes.length;
+            if (length == 0) return 0;
+
+            uint256 ynethRate = IERC4626(MC.YNETH).previewRedeem(1e18);
```
danoctavian: here you should actually use requests[i].redemptionRateAtRequestTime instead of ynETHRate.

this will save all the gas here since you already load the request in memory.

danoctavian: Likewise for ynLSDe, use requests[i]. redemptionRateAtRequestTime

src/BaseVault.sol
```diff
@@ -0,0 +1,641 @@
...
+        emit Pause(false);
+    }
+
+    /**
+     * @notice Processes the accounting of the vault by calculating the total base balance.
+     * @dev This function iterates through the list of assets, gets their balances and rates,
+     *      and updates the total assets denominated in the base asset.
+     */
+    function processAccounting() public virtual {
+        uint256 totalBaseBalance = address(this).balance;
```
danoctavian, line 586: we need to strip this out for the BTC and USD versions

PR #7, opened 2024-10-18
Feature: ETH starter vault

xhad:
This converts the SingleVault to a WETH 4626 vault.

Notes:

1. Adds a fallback function to convert eth to WETH and deposit for shares from direct transfers
2. Remove all of the timelock controller for function admin
3. Updated tests, but still need to do a few more


src/SingleVault.sol
```diff
@@ -1,99 +1,64 @@
...
+            _mintSharesForETH(msg.value);
+        }
     }
 
-    function _retrieveERC4626Storage() internal pure returns (ERC4626Storage storage $) {
-        assembly {
-            $.slot := ERC4626StorageLocation
+    fallback() external payable nonReentrant {
+        if (msg.value > 0) {
+            _mintSharesForETH(msg.value);
```
danoctavian, line 34: this implementation assumes the rate of ETH:Shares is 1:1

Which it may not be even for the starter vault.

For example, if i send WETH directly to the vault, it will increase the rate, as ERC4626Upgradeable counts totalAssets using `$._asset.balanceOf(address(this))`

the easiest implementation here is probably using the _mint function, and previewDeposit.

Have to be careful to mint WETH *only after* calling previewDeposit to not inflate totalAssets before it's processed

src/SingleVault.sol
```diff
@@ -1,99 +1,64 @@
...
-import {ISingleVault} from "src/ISingleVault.sol";
+import {ISingleVault} from "src/interface/ISingleVault.sol";
 
-contract SingleVault is ISingleVault, ERC4626Upgradeable, TimelockControllerUpgradeable, ReentrancyGuardUpgradeable {
-    using Math for uint256;
+/* ynETH Pre-Launch Vault */
 
-    bytes32 private constant ERC4626StorageLocation = 0x0773e532dfede91f04b12a73d3d2acd361424f41f76b4fb79f090161e36b4e00;
-    IStakeManager private constant stakeManager = IStakeManager(0x1adB950d8bB3dA4bE104211D5AB038628e477fE6);
+contract SingleVault is ISingleVault, ERC4626Upgradeable, AccessControlUpgradeable, ReentrancyGuardUpgradeable {
```
danoctavian, line 10: the unstated assumption for this implementation is that whatever we pick for 

asset() is 1:1 to ETH, which is fine

src/SingleVault.sol
```diff
@@ -1,99 +1,64 @@
...
-        if (asset_ == IERC20(address(0))) revert AssetZeroAddress();
-        if (bytes(name_).length == 0) revert NameEmpty();
-        if (bytes(symbol_).length == 0) revert SymbolEmpty();
-        if (admin_ == address(0)) revert AdminZeroAddress();
-        if (proposers_.length == 0) revert ProposersEmpty();
-        if (executors_.length == 0) revert ExecutorsEmpty();
+    receive() external payable nonReentrant {
+        if (msg.value > 0) {
+            _mintSharesForETH(msg.value);
+        }
```
danoctavian, line 29: can we emit an additional event here for the receive() call?

danoctavian: this implementation assumes the rate of ETH:Shares is 1:1

Which it may not be even for the starter vault.

For example, if i send WETH directly to the vault, it will increase the rate, as ERC4626Upgradeable counts totalAssets using $._asset.balanceOf(address(this))

the easiest implementation here is probably using the _mint function, and previewDeposit, as inspired by the current function deposit function.

Have to be careful to mint WETH only after calling previewDeposit to not inflate totalAssets before it's processed

src/SingleVault.sol
```diff
@@ -1,99 +1,64 @@
...
     }
 
-    function _retrieveERC4626Storage() internal pure returns (ERC4626Storage storage $) {
-        assembly {
-            $.slot := ERC4626StorageLocation
+    fallback() external payable nonReentrant {
```
danoctavian, line 32: if receive() is defined, fallback() is not needed

receive() handles all direct ETH deposits.

fallback() is then meant for unexpected callss and calldata and we don't have that use case here

src/SingleVault.sol
xhad: @danoctavian so you're saying the underlying asset of the ETH vault is wstETH?

danoctavian: no, i's not, it will be WETH

the implementation of receive() actually clears the fact that asset() must be WETH so it's clear asset() = WETH

src/SingleVault.sol
```diff
@@ -1,99 +1,62 @@
...
+            _mintSharesForETH(msg.value);
         }
     }
 
-    function totalAssets() public view override(ERC4626Upgradeable, IERC4626) returns (uint256) {
-        ERC4626Storage storage $ = _retrieveERC4626Storage();
-        uint256 assetBalance = $._asset.balanceOf(address(this));
-        return _convertSnBnbToBnb(assetBalance);
-    }
+    function _mintSharesForETH(uint256 amount) private {
```
danoctavian, line 32: to make it consistent with how deposit behaves, we should check for maxDeposit

```
        uint256 maxAssets = maxDeposit(receiver);
        if (assets > maxAssets) {
            revert ERC4626ExceededMaxDeposit(receiver, assets, maxAssets);
        }
```

danoctavian:
other than the maxDeposit check, it looks correct now

src/SingleVault.sol
xhad: changed my mind here and added this check, thanks
7856afcf6858ee0e9c9ead76e6105e1326e6b204

PR #12, opened 2024-11-22
Fix/mainnet tests

xhad:
This adds a threshold asertion helper to fix 1 wei variance in rounding. 

It also updates the invariant tests for mainnet

### DIFF
--- a/src/Vault.sol
+++ b/src/Vault.sol
@@ -141,7 +141,7 @@ contract Vault is IVault, ERC20PermitUpgradeable, AccessControlUpgradeable, Reen
         }
 
         uint256 ownerShares = balanceOf(owner);
-        (uint256 maxAssets,) = _convertToAssets(asset(), ownerShares, Math.Rounding.Floor);
+        uint256 maxAssets = convertToAssets(ownerShares);
 
         return bufferAssets < maxAssets ? bufferAssets : maxAssets;
     }
@@ -162,10 +162,7 @@ contract Vault is IVault, ERC20PermitUpgradeable, AccessControlUpgradeable, Reen
         }
 
         uint256 ownerShares = balanceOf(owner);
-        (uint256 ownerAssets,) = _convertToAssets(asset(), ownerShares, Math.Rounding.Floor);
-        (uint256 maxShares,) = _convertToShares(asset(), bufferAssets, Math.Rounding.Floor);
-
-        return bufferAssets < ownerAssets ? maxShares : ownerShares;
+        return bufferAssets < previewRedeem(ownerShares) ? previewWithdraw(bufferAssets) : ownerShares;
     }
 
     /**
@@ -193,7 +190,7 @@ contract Vault is IVault, ERC20PermitUpgradeable, AccessControlUpgradeable, Reen
         if (paused()) {
             revert Paused();
         }
-        (uint256 assets, uint256 baseAssets) = _convertToAssets(asset(), shares, Math.Rounding.Ceil);
+        (uint256 assets, uint256 baseAssets) = _convertToAssets(asset(), shares, Math.Rounding.Floor);
         _deposit(asset(), _msgSender(), receiver, assets, shares, baseAssets);
         return assets;
     }
--- a/src/Vault.sol
+++ b/src/Vault.sol
@@ -420,7 +420,7 @@ contract Vault is IVault, ERC20PermitUpgradeable, AccessControlUpgradeable, Reen
     }
 
     /**
-     * @notice Internal function to convert shares to assets.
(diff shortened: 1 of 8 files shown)


src/Vault.sol
```diff
@@ -458,7 +452,7 @@ contract Vault is IVault, ERC20PermitUpgradeable, AccessControlUpgradeable, Reen
         returns (uint256, uint256)
```
danoctavian, line 449: Related to the other computation

https://github.com/yieldnest/yieldnest-vault/blob/217749891802a3a459d81e5c9dc9941153f8e8d9/src/Vault.sol#L437

_convertToAssets doesn't seem correct as it would first need to convert the shares to base asset, and then convert that to asset_.

Right now the conversion `_convertAssetToBase(asset_, shares);` converts shares as though they are assets

danoctavian: i didn't test this but do let me know if i'm off here

 function _convertToAssets(address asset_, uint256 shares, Math.Rounding rounding)
 
 I would expect this to first convert the shares to base asset, and then covert the base asset to asset_

src/Vault.sol
```diff
@@ -458,7 +452,7 @@ contract Vault is IVault, ERC20PermitUpgradeable, AccessControlUpgradeable, Reen
...
     {
         uint256 baseAssets = _convertAssetToBase(asset_, assets);
-        uint256 shares = baseAssets.mulDiv(totalAssets() + 1, totalSupply() + 10 ** 0, rounding);
+        uint256 shares = baseAssets.mulDiv(totalSupply() + 10 ** 0, totalAssets() + 1, rounding);
```
danoctavian, line 452: this fix looks correct as in it should be 

BaseAssets * totalSupply / totalAssets and decimal offset is hardcoded to zero here

src/Vault.sol
xhad: @danoctavian great, thanks for noticing this.
I pushed an update here kindly review when you can.

src/Vault.sol
```diff
@@ -431,8 +431,8 @@ contract Vault is IVault, ERC20PermitUpgradeable, AccessControlUpgradeable, Reen
         view
         returns (uint256, uint256)
     {
-        uint256 baseAssets = _convertAssetToBase(asset_, shares);
-        uint256 assets = baseAssets.mulDiv(totalAssets() + 1, totalSupply() + 10 ** 0, rounding);
+        uint256 assets = shares.mulDiv(totalAssets() + 1, totalSupply() + 10 ** 0, rounding);
+        uint256 baseAssets = _convertAssetToBase(asset_, assets);
```
xhad, line 435: @danoctavian another approach that might be better is to add the reverse of _convertAssetToBase....

_convertBaseToAsset(_asset, shares);

It seems like converting the assets here, which might be any given _asset, is wrong.

src/module/ETHRates.sol
```diff
@@ -78,7 +78,7 @@ contract ETHRates is IRateProvider {
     }
 
     function _getStETHRate() internal view returns (uint256) {
-        return IStETH(STETH).getSharesByPooledEth(1e18);
+        return IStETH(STETH).getPooledEthByShares(1e18);
```
danoctavian, line 83: this looks correct now, matches what we have for ynEigen

https://github.com/yieldnest/yieldnest-protocol/blob/main/src/ynEIGEN/LSDRateProvider.sol#L52

src/Vault.sol
danoctavian: Yes that is correct we need `_convertBaseToAsset ` 

In words:
 first convert the shares to base asset
 
then covert the base asset to asset_. This last step is achieved with `_convertBaseToAsset `

danoctavian: One issue here why this wasn't caught i believe is because we don't have fuzz testing for  this

I'd consider making a public `    function _convertToAssets(address asset_, uint256 shares, Math.Rounding rounding)` function and fuzz test that

Or a version of function mint, that also specifies the asset, to be equivalent to depositsset

xhad: There is fuzz testing for this, but I think it needs to be tested via scenario based testing that simulates accounting flows.

https://github.com/yieldnest/yieldnest-vault/blob/cfdec9e0c6a30a08e5d1c34ae071374f0bc5606c/test/mainnet/invariants.spec.sol#L202

danoctavian: The converToAssets hardcodes asset_ to asset(), which is base asset hence why it doesn't test the whole functionality of `function _convertToAssets(address asset_, uint256 shares, Math.Rounding rounding)`

xhad: Will open a new PR to add mintAsset and previewMintAsset

PR #18, opened 2024-11-25
Feature:  Strategy contract

xhad:
Adds a generic strategy vault.

1. Moves the Vault logic to an abstract BaseVault contract and create the Vault and Strategy contracts on top.


src/Strategy.sol
```diff
@@ -0,0 +1,616 @@
...
+
+        uint256 shares = previewDeposit(amount);
+        _mint(sender, shares);
+        emit Deposit(sender, sender, amount, shares);
+    }
+
+    /**
+     * @notice Fallback function to handle native asset transfers.
+     */
+    receive() external payable {
```
danoctavian: i'd remove any specific actions on receive, just leave it empty

src/Strategy.sol
```diff
@@ -0,0 +1,616 @@
...
+    ReentrancyGuardUpgradeable,
+    SafeERC20
+} from "./Common.sol";
+
+import {IVault} from "src/interface/IVault.sol";
+import {IStrategy} from "src/interface/IStrategy.sol";
+import {IProvider} from "src/interface/IProvider.sol";
+import {Guard} from "src/module/Guard.sol";
+
+contract Strategy is IVault, ERC20PermitUpgradeable, AccessControlUpgradeable, ReentrancyGuardUpgradeable {
```
danoctavian: since most of the logic is the same as Vault, should we have a BaseVault 

that they both inherit?

src/Strategy.sol
xhad: implemented this change

src/Strategy.sol
xhad: created BaseVault

PR #21, opened 2024-11-26
Feature/adjustable decimals

danoctavian:


src/interface/IVault.sol
```diff
@@ -5,15 +5,16 @@ import {IERC4626} from "src/Common.sol";
 
 interface IVault is IERC4626 {
     struct VaultStorage {
-        bool paused;
         uint256 totalAssets;
         address provider;
         address buffer;
+        bool paused;
+        uint8 decimals;
```
danoctavian, line 12: the vault can have a different number of decimals from the underlying primary deposit asset.

This way we can for example have 18 decimals, but still have USDC as a primary deposit asset.

src/BaseVault.sol
```diff
@@ -428,7 +428,7 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
     function _convertAssetToBase(address asset_, uint256 assets) internal view virtual returns (uint256) {
         if (asset_ == address(0)) revert ZeroAddress();
         uint256 rate = IProvider(provider()).getRate(asset_);
-        return assets.mulDiv(rate, 1e18, Math.Rounding.Floor);
+        return assets.mulDiv(rate, 10 ** (_getAssetStorage().assets[asset()].decimals), Math.Rounding.Floor);
```
danoctavian, line 431: this computation assumes that the primary deposit asset has 18 decimals.

that's not the case for a lot of the BTC coins like cbBTC or stablecoins like USDC.

PR #25, opened 2024-11-27
Fix: Testing for provider

xhad:
Adds test, makes small updates to provider

### DIFF
--- a/src/module/Provider.sol
+++ b/src/module/Provider.sol
@@ -28,7 +28,7 @@ contract Provider is IProvider {
         } else if (asset == MC.WSTETH) {
             return IStETH(MC.STETH).getPooledEthByShares(1e18);
         } else if (asset == MC.METH) {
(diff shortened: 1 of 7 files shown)


src/module/Provider.sol
```diff
@@ -12,61 +21,47 @@ import {MainnetContracts as MC} from "script/Contracts.sol";
...
-}
+        if (asset == MC.WSTETH) {
+            return IStETH(MC.STETH).getPooledEthByShares(1e18);
+        }
 
-interface IMETH {
-    function ratio() external view returns (uint256);
-}
+        if (asset == MC.METH) {
+            return IMETH(MC.METH_STAKING_MANAGER).mETHToETH(1e18);
```
danoctavian, line 42: noticed the change here, this looks correct, i had never used the previous one before with ratio()

danoctavian (approved):

PR #28, opened 2024-11-30
add custom validator

danoctavian:
adding custom validator that can override the ruleset when extra validation is needed.

Example: complex struct or array parsing, etc.

This way you can have best of both worlds. The rule based validation for general case which is just address whitelisting for now

and the custom logic validator a la Mellow (eg: https://github.com/mellow-finance/mellow-vaults/blob/main/src/interfaces/validators/IValidator.sol) only when needed.

### DIFF
--- a/test/unit/processor.t.sol
+++ b/test/unit/processor.t.sol
@@ -11,6 +11,20 @@ import {Etches} from "test/unit/helpers/Etches.sol";
 import {WETH9} from "test/unit/mocks/MockWETH.sol";
 import {SetupVault} from "test/unit/helpers/SetupVault.sol";
 import {MockSTETH} from "test/unit/mocks/MockST_ETH.sol";
+import {IValidator} from "src/interface/IValidator.sol";
+
+// Mock validator contract for testing
+contract MockValidator is IValidator {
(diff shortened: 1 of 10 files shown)


test/unit/processor.t.sol
```diff
@@ -238,4 +253,86 @@ contract VaultProcessUnitTest is Test, MainnetActors, Etches {
         vm.expectRevert();
         vault.processor(targets, values, data);
     }
+
+    function _setupValidatorRule() internal returns (MockValidator) {
+        bytes4 funcSig = bytes4(keccak256("deposit2()"));
+
+        IVault.ParamRule[] memory paramRules = new IVault.ParamRule[](2);
+
+        // First param rule for uint256 amount
```
dan13ram, line 262: Param rules are never looked at when a validator is present, so we can just send an empty array here?

dan13ram (approved):
Awesome stuff. Easy to add custom validation if we ever need it

test/unit/processor.t.sol
danoctavian: yeah that is the case

PR #33, opened 2024-12-09
added base vault viewer

dan13ram:


src/BaseVaultViewer.sol
```diff
@@ -0,0 +1,96 @@
...
+                decimals: assetParams.decimals
+            });
+        }
+    }
+
+    function getRate() external view returns (uint256) {
+        IVault vault = IVault(_getStorage().vault);
+        uint256 totalSupply = vault.totalSupply();
+        uint256 totalAssets = vault.totalAssets();
+        if (totalSupply == 0 || totalAssets == 0) return 1 ether;
```
danoctavian: this assumes the denomination asset has 18 decimals therefore the rate is 1e18.

This is not guaranteed, we have not yet but we may for example decide to do a BTC asset with 8 decimals

the rate is is expressed in the denomination set so it should be 10 ** vault.asset().decimals()

src/BaseVaultViewer.sol
```diff
@@ -0,0 +1,96 @@
...
+        return 1 ether * totalAssets / totalSupply;
```
danoctavian: same here assumes decimals

also why  not use math.muldiv ?

src/BaseVaultViewer.sol
```diff
@@ -0,0 +1,96 @@
...
+
+            uint256 assetBalance = asset.balanceOf(address(vault));
+            uint256 baseBalance = Math.mulDiv(assetBalance, rate, 10 ** assetParams.decimals, Math.Rounding.Floor);
+
+            assetsInfo[i] = AssetInfo({
+                asset: assets[i],
+                name: asset.name(),
+                symbol: asset.symbol(),
+                rate: rate,
+                ratioOfTotalAssets: (baseBalance > 0 && totalAssets > 0) ? baseBalance * DECIMALS / totalAssets : 0,
```
dan13ram: @danoctavian should this "DECIMALS" also be the decimals of the base asset?

danoctavian: no that's just a way to define how many decimals that ratio has for precision purposes

That's up to us

danoctavian (approved):

PR #35, opened 2024-12-11
eth-max-vault full code

danoctavian:
Only for review purposes only.

No need to merge.


src/BaseVault.sol
```diff
@@ -0,0 +1,662 @@
...
+     * @param owner The address of the owner.
+     * @return shares The equivalent amount of shares.
+     */
+    function withdraw(uint256 assets, address receiver, address owner)
+        public
+        virtual
+        nonReentrant
+        returns (uint256 shares)
+    {
+        if (paused()) {
```
johnnyonline, line 224: redundant call, we already check this at `maxWithdraw` below

src/BaseVault.sol
```diff
@@ -0,0 +1,662 @@
...
+     * @param owner The address of the owner.
+     * @return assets The equivalent amount of assets.
+     */
+    function redeem(uint256 shares, address receiver, address owner)
+        public
+        virtual
+        nonReentrant
+        returns (uint256 assets)
+    {
+        if (paused()) {
```
johnnyonline, line 249: redundant call, we already check this at `maxRedeem` below

src/BaseVault.sol
```diff
@@ -0,0 +1,662 @@
...
+    /**
+     * @notice Internal function to convert an asset amount to base denomination.
+     * @param asset_ The address of the asset.
+     * @param assets The amount of the asset.
+     * @return uint256 The equivalent amount in base denomination.
+     */
+    function _convertAssetToBase(address asset_, uint256 assets) internal view virtual returns (uint256) {
+        if (asset_ == address(0)) revert ZeroAddress();
+        uint256 rate = IProvider(provider()).getRate(asset_);
+        return assets.mulDiv(rate, 10 ** (_getAssetStorage().assets[asset()].decimals), Math.Rounding.Floor);
```
johnnyonline: what happens if there's an asset with different decimals from `assets[asset()].decimals`?

same in `_convertBaseToAsset`

johnnyonline: also i wonder if when exiting the vault we should still round down

src/BaseVault.sol
```diff
@@ -0,0 +1,662 @@
...
+
+    /**
+     * @notice Processes the accounting of the vault by calculating the total base balance.
+     * @dev This function iterates through the list of assets, gets their balances and rates,
+     *      and updates the total assets denominated in the base asset.
+     */
+    function processAccounting() public virtual {
+        VaultStorage storage vaultStorage = _getVaultStorage();
+
+        uint256 totalBaseBalance = vaultStorage.countNativeAsset ? address(this).balance : 0;
```
johnnyonline: seems we cant set `vaultStorage.countNativeAsset` anywhere? if it cant be changed why not make it an immutable var and save the storage read?

johnnyonline: also why do we have this `address(this).balance` at all? seems cant deposit/withdraw native eth

src/BaseVault.sol
```diff
@@ -0,0 +1,662 @@
...
+
+        AssetStorage storage assetStorage = _getAssetStorage();
+        address[] memory assetList = assetStorage.list;
+        uint256 assetListLength = assetList.length;
+        uint256 baseAssetUnit = 10 ** (assetStorage.assets[asset()].decimals);
```
johnnyonline: `assetStorage.assets[asset()].decimals` -- what happens if we have an asset with different decimals?

src/BaseVault.sol
```diff
@@ -0,0 +1,662 @@
...
+
+        for (uint256 i = 0; i < assetListLength; i++) {
+            uint256 balance = IERC20(assetList[i]).balanceOf(address(this));
+            if (balance == 0) continue;
+            uint256 rate = IProvider(provider()).getRate(assetList[i]);
+            totalBaseBalance += balance.mulDiv(rate, baseAssetUnit, Math.Rounding.Floor);
+        }
+
+        _getVaultStorage().totalAssets = totalBaseBalance;
+        emit ProcessAccounting(block.timestamp, totalBaseBalance);
```
johnnyonline: do we need to emit `block.timestamp`? seems like redundant cost?

src/BaseVault.sol
```diff
@@ -0,0 +1,662 @@
...
+        vaultStorage.paused = false;
+        emit Pause(false);
+    }
+
+    /**
+     * @notice Processes the accounting of the vault by calculating the total base balance.
+     * @dev This function iterates through the list of assets, gets their balances and rates,
+     *      and updates the total assets denominated in the base asset.
+     */
+    function processAccounting() public virtual {
```
johnnyonline: how do we count the assets that are invested in other strategies? i think im missing something basic here

src/BaseVault.sol
```diff
@@ -0,0 +1,662 @@
...
+    }
+
+    /**
+     * @notice Processes a series of calls to target contracts.
+     * @param targets The addresses of the target contracts.
+     * @param values The values to send with the calls.
+     * @param data The calldata for the calls.
+     * @return returnData The return data from the calls.
+     */
+    function processor(address[] calldata targets, uint256[] memory values, bytes[] calldata data)
```
johnnyonline, line 725: seems that owners can rug the vault with this (even though `guard`). probably not ideal?

src/BaseVault.sol
```diff
@@ -0,0 +1,662 @@
...
+    }
+
+    constructor() {
+        _disableInitializers();
+    }
+
+    /**
+     * @notice Fallback function to handle native asset transfers.
+     */
+    receive() external payable {
```
johnnyonline, line 753: why do we need that? i dont see a way to withdraw eth from here too

also donation attack?

src/Strategy.sol
```diff
@@ -0,0 +1,121 @@
...
+
+contract Strategy is BaseVault {
+    /**
+     * @notice Initializes the Strategy Vault.
+     * @param admin The address of the admin.
+     * @param name The name of the vault.
+     * @param symbol The symbol of the vault.
+     * @param decimals_ The number of decimals for the vault token.
+     */
+    function initialize(address admin, string memory name, string memory symbol, uint8 decimals_)
```
johnnyonline, line 15: we init the decimals of the token but not the token? or am i missing it?

src/Strategy.sol
```diff
@@ -0,0 +1,121 @@
...
+     * @param asset_ The address of the asset.
+     * @param caller The address of the caller.
+     * @param receiver The address of the receiver.
+     * @param assets The amount of assets to deposit.
+     * @param shares The amount of shares to mint.
+     * @param baseAssets The base asset convertion of shares.
+     * @dev This is an example:
+     *     The _deposit function for strategies needs an override
+     */
+    function _deposit(
```
johnnyonline, line 71: why do we override this as well and remove the asset check?

src/Vault.sol
```diff
@@ -0,0 +1,101 @@
...
+        bool countNativeAsset_
+    ) external virtual initializer {
+        __ERC20_init(name, symbol);
+        __AccessControl_init();
+        __ReentrancyGuard_init();
+        _grantRole(DEFAULT_ADMIN_ROLE, admin);
+
+        VaultStorage storage vaultStorage = _getVaultStorage();
+        vaultStorage.paused = true;
+        vaultStorage.decimals = decimals_;
```
johnnyonline, line 62: i got to be missing the asset initialization

test/mainnet/curve.spec.sol
```diff
@@ -0,0 +1,229 @@
...
+            callData[1] = exchangeData;
+
+            vm.startPrank(PROCESSOR);
+            vault.processor(targets, values, callData);
+            vm.stopPrank();
+        }
+
+        // Assert stETH balance is 0 and ETH balance matches expected amount
+        assertApproxEqAbs(IERC20(MC.STETH).balanceOf(address(vault)), amount - swapAmount, 2);
+        assertApproxEqAbs(address(vault).balance, minOut, 2);
```
johnnyonline, line 175: seems this eth is stuck in the vault?

test/mainnet/curve.spec.sol
```diff
@@ -0,0 +1,229 @@
...
+        vault.processAccounting();
+
+        // Assert total assets remains unchanged after swap
+        assertApproxEqAbs(vault.totalAssets(), totalAssetsBefore, delta);
+    }
+
+    function test_Vault_Curve_swapETHToStETH() public {
+        uint256 amount = 100 ether;
+
+        // User deposits ETH
```
johnnyonline, line 186: user gets shares here or this is a donation?

src/BaseVault.sol
```diff
@@ -0,0 +1,662 @@
...
+
+        uint256 totalBaseBalance = vaultStorage.countNativeAsset ? address(this).balance : 0;
+
+        AssetStorage storage assetStorage = _getAssetStorage();
+        address[] memory assetList = assetStorage.list;
+        uint256 assetListLength = assetList.length;
+        uint256 baseAssetUnit = 10 ** (assetStorage.assets[asset()].decimals);
+
+        for (uint256 i = 0; i < assetListLength; i++) {
+            uint256 balance = IERC20(assetList[i]).balanceOf(address(this));
```
johnnyonline, line 712: donation attack?

src/BaseVault.sol
```diff
@@ -0,0 +1,662 @@
...
+        _getProcessorStorage().rules[target][functionSig] = rule;
+        emit SetProcessorRule(target, functionSig, rule);
+    }
+
+    /**
+     * @notice Adds a new asset to the vault.
+     * @param asset_ The address of the asset.
+     * @param active_ Whether the asset is active or not.
+     */
+    function addAsset(address asset_, bool active_) public virtual onlyRole(ASSET_MANAGER_ROLE) {
```
johnnyonline, line 580: the way of adding strategies (as assets - meaning we'll need to wrap every strategy?) seems not scaleable as we'll need to iterate over that list when updating total assets. also cant remove strategies?

danoctavian: yeah removal needs to be added exactly

Right now, you can only toggle off the active boolean

src/BaseVault.sol
johnnyonline: also on `processAccounting` -- when are we planning to call this? and should we continuously call this for every vault/asset/strategy (not sure what's the right terminology here) -- is it realistic both from a gas perspective and user security perspective?

src/BaseVault.sol
danoctavian: good point about the gas issue with addAsset should be acceptable for the start, as we'll have on the order of 10 strats at most

src/BaseVault.sol
danoctavian: yeah, we decided to accept this trade-off to simplify balance tracking.

ERC4626 by OZ makes the same tradeoff.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC4626.sol#L117

As we've experienced with ynEigen, dealing with rebasing assets can become very cumbersome

test/mainnet/curve.spec.sol
danoctavian: you can process it with a call to function processor(), wrap it as WETH, deposit it somewhere else etc.

src/BaseVault.sol
danoctavian: good point

src/BaseVault.sol
danoctavian: good catch this is an issue

src/BaseVault.sol
danoctavian: yeah great, question, on the L2s/binance, we can call it quite frequently

on L1 we're thinking 1-2 times a day. 

One question here is we should make it permissioned the way yearn v3 has process_report permissioned.

We investigated an approach that uses the debt-style accounting of yearn v3 allocator to have a processAccounting with an `address asset` parameter that updates only specific assets/strategies, which would improve the gas problem.

However the complexity of that bubbles up in the `function processor` as that would need to update the debt recrod for the assets affected by said processor transaction, which may be 1, 2 or more (eg. you withdraw an tripool LP position from curve into 3 assets, that's 4 assets that need updating)

We decided therefore not to go for this now given the current launch timeline

src/BaseVault.sol
danoctavian: it is, but the maxWithdraw does not revert it simply returns 0, this one throws an explicit error that states it's paused

src/BaseVault.sol
danoctavian: same as above

src/Strategy.sol
danoctavian: this was an example Strategy.sol, it didn't go into audit, we can improve it

src/Vault.sol
danoctavian: you're right we separated it into a setter we should always bootstrap with 1 asset

src/BaseVault.sol
johnnyonline: wouldnt it revert here (https://github.com/yieldnest/yieldnest-vault/blob/eth-max-vault/src/BaseVault.sol#L225) `if ((assets > maxAssets) {revert ExceededMaxWithdraw(owner, assets, maxAssets);}`)?

PR #38, opened 2024-12-13
added getUnderlyingAssets

dan13ram:


src/utils/BaseVaultViewer.sol
```diff
@@ -72,6 +72,14 @@ contract BaseVaultViewer is Initializable {
      * @return assetsInfo An array of AssetInfo structs containing detailed information about each asset
      */
     function getAssets() external view returns (AssetInfo[] memory assetsInfo) {
+        return _getAssets();
+    }
+
+    function getUnderlyingAssets() external view virtual returns (AssetInfo[] memory assetsInfo) {
```
dan13ram, line 79: Only this function is virtual since it will be overridden.

PR #41, opened 2024-12-18
Feat/strategy referral adapter

MrDeadCe11:
adding referral adapter

### DIFF
--- a/src/utils/XReferralAdapter.sol
+++ b/src/utils/XReferralAdapter.sol
@@ -4,7 +4,13 @@ pragma solidity ^0.8.24;
 
 import {IVault} from "src/BaseVault.sol";
 
+interface IKernelStrategy {
+   function getHasAllocator() external view returns (bool hasAllocators) ;
+}
+
 contract XReferralAdapter {
+    /// @notice Role for allocator permissions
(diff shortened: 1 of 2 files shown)


src/utils/XReferralAdapter.sol
```diff
@@ -0,0 +1,92 @@
...
+pragma solidity ^0.8.24;
+
+import {IVault} from "src/BaseVault.sol";
+
+interface IKernelStrategy {
+    function getHasAllocator() external view returns (bool hasAllocators);
+    function hasRole(bytes32 role, address account) external view returns (bool hasRole);
+}
+
+contract XReferralAdapter {
```
danoctavian, line 7: can we call it VaultReferralAdapter ? 

this sounds cooler, but i want to keep it more standard

src/utils/XReferralAdapter.sol
```diff
@@ -0,0 +1,76 @@
...
+        uint256 shares,
+        uint256 timestamp
+    );
+
+    error InvalidVault(address);
+    error ZeroAmount();
+    error ZeroAddress();
+    error SelfReferral();
+
+    constructor() public {}
```
danoctavian, line 28: disable  initializers

PR #48, opened 2024-12-21
fixed spec tests for ynbnbx

dan13ram:


test/mainnet/ynbnbk.spec.sol
```diff
@@ -79,15 +74,62 @@ contract VaultMainnetYnBNBkTest is Test, AssertUtils, MainnetActors {
...
+        uint256 previewAssets = Vault(payable(MC.YNBNBk)).previewRedeem(actualShares);
+
+        assertApproxEqAbs(
+            newTotalAssets,
+            previousTotalAssets + previewAssets,
+            1000,
+            "New total assets should equal to previous total assets plus previewAssets"
+        );
+
+        // TODO: this should be true
```
dan13ram, line 126: @danoctavian Could you take a closer look at this test? 

The provider returns different values for rates for `ynBNBk` and `slisBNB`, and so the totalAssets value changes upon deposit into ynBNBk from ynBNBx. 

We may need to introduce a fix if you think this is important.

danoctavian: let's fix this, it's normaly ynBNBk is not 1:1 with slisBNB, it should use its own rate.

ynBNBk is a yield token that grows over time

danoctavian: this is pre-upgrade as discussed

danoctavian (approved):

PR #51, opened 2024-12-24
fix audt issues

danoctavian:


src/BaseVault.sol
```diff
@@ -154,7 +154,7 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
         }
 
         uint256 ownerShares = balanceOf(owner);
-        uint256 maxAssets = convertToAssets(ownerShares);
+        uint256 maxAssets = previewRedeem(ownerShares);
```
danoctavian, line 157: this way maxWithdraw accounts for fees.
From ERC4626:
„MUST return the maximum amount of assets that could be transferred from owner through withdraw and not cause a revert, which MUST NOT be higher than the actual maximum that would be accepted (it should underestimate if necessary).”

src/BaseVault.sol
```diff
@@ -420,9 +421,11 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
             _spendAllowance(owner, caller, shares);
         }
 
+        // NOTE: burn shares before withdrawing the assets
+        _burn(owner, shares);
```
danoctavian, line 425: implement the Checks-Effects-Interactions pattern.

src/BaseVault.sol
```diff
@@ -674,7 +677,11 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
      * @dev This function iterates through the list of assets, gets their balances and rates,
      *      and updates the total assets denominated in the base asset.
      */
-    function processAccounting() public virtual {
+    function processAccounting() public virtual nonReentrant {
```
danoctavian, line 680: add nonReentrant as well so an evil token cannot reenter through processAccounting while _withdraw is happening

dan13ram (approved):

src/BaseVault.sol
dan13ram: Got it. `KernelStrategy` currently doesn't satisfy this condition, and thats why we use the viewer instead to get maxWithdraw :)

danoctavian: yes, we're ok with that for now there

danoctavian: yes for now

PR #53, opened 2024-12-24
added verify script for ynbnbx

dan13ram:


script/VaultUtils.sol
```diff
@@ -80,15 +80,12 @@ contract VaultUtils {
         setWithdrawAssetRule(vault_, contractAddress, allowList);
     }
 
-    function setWithdrawAssetRule(IVault vault_, address contractAddress, address[] memory assetList) internal {
-        address[] memory allowList = new address[](1);
-        allowList[0] = address(vault_);
-
+    function setWithdrawAssetRule(IVault vault_, address contractAddress, address[] memory allowList) internal {
```
dan13ram: This change is incorrect I think. Only the first param should be assetList the rest should be only the `vault` since on the vault is the owner and receiver is all cases.

PR #54, opened 2024-12-24
Added max vault viewer

dan13ram:


src/utils/MaxVaultViewer.sol
```diff
@@ -0,0 +1,96 @@
...
+            if (assets[i] == asset_) {
+                found = true;
+            }
+        }
+
+        if (!found) revert InvalidAssetAdd(asset_);
+        _;
+    }
+
+    function addKnownAsset(address asset_) external onlyRole(UPDATER_ROLE) onlyVaultAsset(asset_) {
```
danoctavian: let's make this be internal and we have a public one that takes an array

danoctavian: let's call it addUnderlyingAssets

_addUnderlyingAsset for the internal one

src/utils/MaxVaultViewer.sol
```diff
@@ -0,0 +1,96 @@
...
+        if (asset_ == address(0)) revert ZeroAddress();
+        if (_getAssetStorage().assets[asset_]) revert InvalidAssetAdd(asset_);
+
+        _getAssetStorage().assets[asset_] = true;
+        _getAssetStorage().assetsLength += 1;
+    }
+
+    function removeKnownAsset(address asset_) external onlyRole(UPDATER_ROLE) {
```
danoctavian: call it removeUnderlyingAssets

and respectively _addUnderlyingAsset for the internal one

analog to the add

src/utils/MaxVaultViewer.sol
```diff
@@ -0,0 +1,96 @@
...
+        if (allAssets.length < assetsLength) revert InvalidAssets();
+
+        uint256 strategiesLength = allAssets.length - assetsLength;
+
+        address[] memory strategies = new address[](strategiesLength);
+        uint256[] memory balances = new uint256[](strategiesLength);
+
+        for (uint256 i = 0; i < strategiesLength; ++i) {
+            if (_getAssetStorage().assets[allAssets[i]]) continue;
+            strategies[i] = allAssets[i];
```
danoctavian: this now works the opposite way it should work

the strategies are the things that are *not* underyling assets

dan13ram: @danoctavian Here its exactly as you described it, strategies are indeed NOT underlying assets. 

If (asset is underlying) continue
Else add to strategies list

danoctavian: i see it now, please for style purposes let's just do if (! ... the rule) {
}
the continue with {} is hard to read

dan13ram: fixed in [df7ca37](https://github.com/yieldnest/yieldnest-vault/pull/54/commits/df7ca37f9c0167de7115697aedeae9efadd02ba6)

src/utils/MaxVaultViewer.sol
dan13ram: fixed in [df7ca37](https://github.com/yieldnest/yieldnest-vault/pull/54/commits/df7ca37f9c0167de7115697aedeae9efadd02ba6)

src/utils/MaxVaultViewer.sol
dan13ram: fixed in [df7ca37](https://github.com/yieldnest/yieldnest-vault/pull/54/commits/df7ca37f9c0167de7115697aedeae9efadd02ba6)

PR #61, opened 2024-12-26
fixed approval rules and roles in the deploy script

dan13ram:


script/DeployMaxVault.s.sol
```diff
@@ -139,7 +138,7 @@ contract DeployMaxVault is BaseScript {
             assets[2] = contracts.BNBX();
             setDepositAssetRule(vault, contracts.YNBNBK(), assets);
             setWithdrawAssetRule(vault, contracts.YNBNBK(), assets);
-            setApprovalRule(vault, contracts.YNBNBK(), assets);
+            // TODO: fix approval rule for ynbnbk
```
dan13ram: I didn't want to complicate this script since we will be deploying without ynbnbk anyway and configure it separately.

script/DeployMaxVault.s.sol
```diff
@@ -148,7 +147,17 @@ contract DeployMaxVault is BaseScript {
...
+
+        if (contracts.YNWBNBK() != address(0x0b) && contracts.YNCLISBNBK() != address(0x0c)) {
+            address[] memory underlyingVaults = new address[](2);
+            underlyingVaults[0] = contracts.YNWBNBK();
+            underlyingVaults[1] = contracts.YNCLISBNBK();
+            setApprovalRule(vault, contracts.WBNB(), underlyingVaults);
+        } else if (contracts.YNWBNBK() != address(0x0b)) {
+            setApprovalRule(vault, contracts.WBNB(), contracts.YNWBNBK());
+        } else if (contracts.YNCLISBNBK() != address(0x0c)) {
+            setApprovalRule(vault, contracts.WBNB(), contracts.YNCLISBNBK());
```
dan13ram: We need to set a single approval rule, since the rules are stored per contractAddress per funcSig, so only a single rule can be set for `contracts.WBNB()` and `approve` funcSig

script/DeployMaxVault.s.sol
```diff
@@ -148,7 +147,17 @@ contract DeployMaxVault is BaseScript {
             setWithdrawRule(vault, contracts.YNCLISBNBK());
             setDepositAssetRule(vault, contracts.YNCLISBNBK(), contracts.WBNB());
             setWithdrawAssetRule(vault, contracts.YNCLISBNBK(), contracts.WBNB());
-            setApprovalRule(vault, contracts.YNCLISBNBK(), contracts.WBNB());
+        }
+
+        if (contracts.YNWBNBK() != address(0x0b) && contracts.YNCLISBNBK() != address(0x0c)) {
```
danoctavian: let's revert if YWBNBk is 0x0b

we need the buffer to get it going

script/DeployMaxVault.s.sol
```diff
@@ -148,7 +147,17 @@ contract DeployMaxVault is BaseScript {
...
+            address[] memory underlyingVaults = new address[](2);
+            underlyingVaults[0] = contracts.YNWBNBK();
+            underlyingVaults[1] = contracts.YNCLISBNBK();
+            setApprovalRule(vault, contracts.WBNB(), underlyingVaults);
```
danoctavian, line 148: instead of elseif elseif 

Once you get past the revert you have only one if that sets to 2 items or just one the underlyingVaults array



after if, call underlyingVaults

PR #68, opened 2025-01-02
fix issue YINE-c7850b7c-H01

danoctavian:
processAccounting currently assumes the native asset (eg. ETH or BNB) has the same number of decimals as .asset())

This enforces that in _addAsset.


johnnyonline (approved):
lgtm

PR #71, opened 2025-01-09
add a withdrawer demo

danoctavian:


src/base/BaseSyncStrategy.sol
```diff
@@ -0,0 +1,158 @@
...
+import {IERC20, Math, SafeERC20} from "src/Common.sol";
+import {BaseStrategy} from "src/base/BaseStrategy.sol";
+
+/**
+ * @title BaseSyncStrategy
+ * @author Yieldnest
+ * @notice This contract is a base strategy for any underlying protocol with syncDeposit and syncWithdraw flags.
+ * vault.
+ */
+abstract contract BaseSyncStrategy is BaseStrategy {
```
danoctavian, line 13: i would not include this one

src/base/BaseStrategy.sol
```diff
@@ -0,0 +1,396 @@
...
+import {IERC20, Math, SafeERC20} from "src/Common.sol";
+import {Vault} from "src/Vault.sol";
+
+/**
+ * @title BaseStrategy
+ * @author Yieldnest
+ * @notice This contract is a base strategy for any underlying protocol.
+ * vault.
+ */
+abstract contract BaseStrategy is Vault {
```
danoctavian, line 13: this should probably not inherit Vault but BaseVault

danoctavian, line 13: let's try an experiment here with external libraries to reduce code size of contract

Maybe a branch off of this branch is best?

src/withdraws/Withdrawer.sol
```diff
@@ -0,0 +1,39 @@
...
+
+import {BaseAsyncWithdrawStrategy} from "src/base/BaseAsyncWithdrawStrategy.sol";
+
+contract Withdrawer is BaseAsyncWithdrawStrategy {
+    /**
+     * @notice Internal function to handle the assets that are in queue for withdrawal.
+     * @param asset_ The address of the asset.
+     * @dev This function should return the amount in base denomination.
+     */
+    function _asyncWithdrawBalance(address asset_)
```
danoctavian, line 18: this looks like the kind of logic that can be separated down into a separate external library 

especially as it will use multiple cases for each asset

src/withdraws/Withdrawer.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity ^0.8.24;
+
+import {BaseVault} from "src/BaseVault.sol";
+import {ISlisBnbStakeManager} from "src/interface/external/lista/ISlisBnbStakeManager.sol";
+import {MainnetContracts} from "script/Contracts.sol";
```
danoctavian, line 6: i have a concern about this MainnetContracts

when you pull it into a dependency it reads all the variables in there and also increases contract size, even though you use 1,2 of them

maybe that is mitigated automatically if we have the logic of `_asyncWithdrawBalance` in a separate external library?

test/mainnet/fork/ynbnbx.spec.sol
```diff
@@ -59,12 +56,19 @@ contract YnBNBxForkTest is Test, MainnetActors, ProxyUtils, VaultUtils {
         assertGt(vault.balanceOf(alice), aliceSharesBefore, "Should have received shares");
 
         // Check vault state after deposit
-        assertEq(vault.totalAssets(), initialTotalAssets + depositAmount, "Total assets should increase by deposit amount");
-        assertEq(vault.totalSupply(), initialTotalSupply + depositAmount, "Total supply should increase by deposit amount");
+        assertEq(
+            vault.totalAssets(), initialTotalAssets + depositAmount, "Total assets should increase by deposit amount"
```
danoctavian, line 60: is this just forge fmt?

src/base/BaseStrategy.sol
dan13ram: @danoctavian i did this so fees are also included here, thats the main difference between BaseVault and Vault right?

src/withdraws/Withdrawer.sol
dan13ram: Yeah we could def move this logic to a separate library or contract

test/mainnet/fork/ynbnbx.spec.sol
dan13ram: Yes that's right.. Only fmt

src/base/BaseStrategy.sol
danoctavian: i'd keep that out not all strategies have fees for withdrawals so we can just inherit BaseVault

src/base/BaseStrategy.sol
```diff
@@ -0,0 +1,396 @@
...
+    /// @notice Role for allocator permissions
+    bytes32 public constant ALLOCATOR_ROLE = keccak256("ALLOCATOR_ROLE");
+
```
danoctavian, line 16: how about we add a STRATEGY_VERSION = "0.1.0"

like here https://github.com/yearn/tokenized-strategy/blob/master/src/TokenizedStrategy.sol#L346

dan13ram:
Closing in favor of #78

PR #75, opened 2025-01-14
Feat/add vault storage lib

MrDeadCe11:


MrDeadCe11:
well, I thought this would help contract sizes for the Vault... turns out it adds 30 bytes to the Vault.sol.

PR #77, opened 2025-01-14
Feat/add vault storage lib

danoctavian:


src/libraries/AsyncWithdrawLib.sol
```diff
@@ -0,0 +1,51 @@
...
+            totalBaseBalance += balanceInBase;
+        }
+    }
+
+    /**
+     * @notice Internal function to handle the assets that are in queue for withdrawal.
+     * @param asset_ The address of the asset.
+     * @dev This function should return the amount in base denomination.
+     */
+    function asyncWithdrawBalance(address asset_) public view returns (uint256 assets, uint256 baseAssets) {
```
dan13ram, line 35: I wonder if we could actually move this logic to the Provider.sol, and we can update it on the contract as we support more assets

dan13ram:
Closing in favor of #78

PR #79, opened 2025-01-19
added ynLSDe and ynETH async withdraw flow

dan13ram:


src/library/AsyncWithdrawLib.sol
```diff
@@ -23,43 +25,57 @@ library AsyncWithdrawLib {
...
 
-    function getAssets(address asset) public pure returns (uint256 assets) {
-        if (asset == MC.WETH) {
-            return 0;
+    function getAssetsQueuedForWithdrawal(address queueManager_, address owner) public view returns (uint256 assets) {
+        IWithdrawalQueueManager queueManager = IWithdrawalQueueManager(queueManager_);
+        (, IWithdrawalQueueManager.WithdrawalRequest[] memory requests) = queueManager.withdrawalRequestsForOwner(owner);
+        for (uint256 i = 0; i < requests.length; i++) {
+            if (!requests[i].processed) {
+                assets += requests[i].amount;
```
danoctavian: there's a gotcha here with the withdrawalqueuemanager

it has its own `redemptionRateAtRequestTime` which is what's actually used.

which means you can't use the current redemption rate of ynETH or ynLSDe, the real redemption rate doesn't increase.

this breaks the above abstraction with the asset amount multiplied by current rate as provided by provider, we'll need to deal with this at this level

https://github.com/yieldnest/yieldnest-protocol/blob/release-candidate/src/WithdrawalQueueManager.sol#L292

In reality it can ever be lower than that if there was a lower rate at "finalization" which may happen if there's slashing.

since we didn't create a good view function for this yet, let's assume for this implementation its redemptionRateAtRequestTime and a NOTE there above the logic above (needs to be fixed - assumes no slashing for now, as in reality eigenlayer slashing is not active yet).

test/mainnet/helpers/SetupWithdrawer.sol
```diff
@@ -0,0 +1,71 @@
...
+
+        // test cannot unpause vault without buffer
+        vm.expectRevert();
+        vault.unpause();
+
+        // set the rate provider contract
+        vault.setProvider(MC.PROVIDER);
+
+        // Add assets: Base asset always first
+        vault.addAsset(MC.WETH, true, true);
```
danoctavian, line 56: let's please write a measurement of the gas profile for processAccounts when the vault is set up like this.

the main optimization here in processAccounting is that if there's no vault balance at all for an asset computation doesn't proceed (don't read the rate)

Please think if there's any gas optimizations here for this async loops, we want to optimize this as much as we can for now

test/mainnet/withdrawer.spec.sol
```diff
@@ -0,0 +1,311 @@
...
+        IWithdrawalQueue queue = IWithdrawalQueue(MC.WSTETH_WQ);
+
+        (uint256 assets,) = vault.asyncWithdrawBalance(asset_);
+        assertEq(assets, 0, "Queued assets should be zero");
+        uint256 totalAssets = vault.totalAssets();
+
+        uint256[] memory amounts = new uint256[](1);
+        amounts[0] = amount;
+
+        // TODO: replace with processor
```
danoctavian: yes please

test/mainnet/withdrawer.spec.sol
```diff
@@ -0,0 +1,311 @@
+// SPDX-License-Identifier: BSD Clause-3
```
danoctavian, line 1: main note here is when you finalize ynETH and ynLSDe withdrawals do not attempt to simulate a full proper withdrawal cycle, not needed, too complex

just donate to the ynETHAssetsRedemptionVault and ssetsRedemptionVault respectively (See spec for details)

be careful that when donations are done, the ETH rate of ynETH and ynLSDe increases

script/Contracts.sol
```diff
@@ -26,8 +26,13 @@ library MainnetContracts {
     address public constant PROVIDER = address(123456789); // TODO: Update with deployed Provider
     address public constant BUFFER = address(987654321); // TODO: Update with deployed buffer
 
-    address public constant YNLSDE_WM = 0x8Face3283E20b19d98a7a132274B69C1304D60b4;
-    address public constant YNETH_WM = 0x0BC9BC81aD379810B36AD5cC95387112990AA67b;
+    address public constant WSTETH_WQ = 0x889edC2eDab5f40e902b864aD4d7AdE8E412F9B1;
+
+    address public constant YNLSDE_WQM = 0x8Face3283E20b19d98a7a132274B69C1304D60b4;
```
danoctavian: lets make these names explicit  - i know they're long but i'd rather not think what they are

src/interface/IWithdrawalQueueManager.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: BSD 3-Clause License
+pragma solidity ^0.8.24;
+
+interface IWithdrawalQueueManager {
```
danoctavian, line 4: good solution now to use interface as opposed to submodules due to remappings.txt hell

src/library/AsyncWithdrawLib.sol
```diff
@@ -23,43 +25,57 @@ library AsyncWithdrawLib {
...
+        // TODO: support WOETH
 
-        if (asset == MC.STETH) {
-            return 0;
+        if (asset == MC.WSTETH) {
+            IWithdrawalQueue queue = IWithdrawalQueue(MC.WSTETH_WQ);
+            uint256[] memory requestIds = queue.getWithdrawalRequests(owner);
+            IWithdrawalQueue.WithdrawalRequestStatus[] memory statuses = queue.getWithdrawalStatus(requestIds);
+            for (uint256 i = 0; i < statuses.length; i++) {
+                assets += statuses[i].amountOfShares;
```
danoctavian: i haven't looked into this in depth, i am curious, while you're in a withdrawal state, do you still earn yield with WSTETH if this is priced in shares then?

because for ynETH and ynLSDe as mentioned below we stop giving you yield

dan13ram: When claiming the share rate is calculated here:

https://github.com/lidofinance/core/blob/901c0e19b752f5ea03118d71881d8b72ccee44c0/contracts/0.8.9/WithdrawalQueueBase.sol#L507

I'm not sure how this maxShareRate is calculated, but it is input from offchain through an oracle to this queue when finalizing a checkpoint.

test/mainnet/withdrawer.spec.sol
dan13ram: Yes we are donating right now to the redemption vaults in the setUp of the tests

src/library/AsyncWithdrawLib.sol
dan13ram: @danoctavian Do we also want to take into account the fees here when calculating the total assets when a withdrawal request is in queue ? 

Currently in order to reduce the complexity here I have moved the fee calculations out of this flow. It is calculated in the tests to ensure that the right fees are removed when the claim happens.

src/library/AsyncWithdrawLib.sol
dan13ram: It looks like it doesn't really matter the share rate, since they also provide amountOfStETH which is basically the underlying assets in base which is what we require. So we don't need to do any extra calculations here.

src/library/AsyncWithdrawalLib.sol
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * @notice function to handle the assets that are in queue for withdrawal.
+     * @param asset_ The address of the asset.
+     * @dev This function should return the amount in base denomination.
+     */
+    function asyncWithdrawalBalance(address asset_, address owner) public view returns (uint256 baseAssets) {
+        baseAssets = _asyncWithdrawalBalance(asset_, owner);
+    }
+
+    function _asyncWithdrawalBalanceYN(address queueManager_, address owner)
```
danoctavian: call it _asyncWithdrawalBalanceYNAsset

src/library/AsyncWithdrawalLib.sol
```diff
@@ -0,0 +1,86 @@
...
+                // NOTE: needs to be fixed - assumes no slashing for now, as in reality eigenlayer slashing is not active yet
+                // also we do not account for the fees here
+                baseAssets += requests[i].amount * requests[i].redemptionRateAtRequestTime / 1e18;
+            }
+        }
+        return baseAssets;
+    }
+
+    function _asyncWithdrawalBalance(address asset, address owner) private view returns (uint256 baseAssets) {
+        // TODO: support WOETH
```
danoctavian: remove TODO now?

src/library/OriginWithdrawalLib.sol
```diff
@@ -0,0 +1,88 @@
...
+        emit WOETHWithdrawalRequested(amount, requestId);
+    }
+
+    function claimWithdrawalsWOETH(uint256[] calldata requestIds)
+        public
+        returns (uint256[] memory amounts, uint256 totalAmount)
+    {
+        IOETHVault oethVault = IOETHVault(MC.OETH_VAULT);
+
+        (amounts, totalAmount) = oethVault.claimWithdrawals(requestIds);
```
danoctavian, line 72: this withdraws OETH not WOETH, so OETH needs listing as asset and a rate

we need to make sure that's in the provider, currently it isn't so it needs to be updated also the vault should have wrap/unwrap rules for WOETH/OETH

https://github.com/yieldnest/yieldnest-vault/blob/eth-max-vault/src/module/Provider.sol

src/library/AsyncWithdrawalLib.sol
dan13ram: fixed in [9873862](https://github.com/yieldnest/yieldnest-vault/pull/79/commits/98738628f3afc044fba386f947886809db694a93)

src/library/AsyncWithdrawalLib.sol
dan13ram: fixed in [9873862](https://github.com/yieldnest/yieldnest-vault/pull/79/commits/98738628f3afc044fba386f947886809db694a93)

src/library/AsyncWithdrawalLib.sol
```diff
@@ -0,0 +1,86 @@
...
+        private
+        view
+        returns (uint256 baseAssets)
+    {
+        IWithdrawalQueueManager queueManager = IWithdrawalQueueManager(queueManager_);
+        (, IWithdrawalQueueManager.WithdrawalRequest[] memory requests) = queueManager.withdrawalRequestsForOwner(owner);
+        for (uint256 i = 0; i < requests.length; i++) {
+            if (!requests[i].processed) {
+                // NOTE: needs to be fixed - assumes no slashing for now, as in reality eigenlayer slashing is not active yet
+                // also we do not account for the fees here
```
danoctavian: please account for fees, the fee info is in the withdrawal request

src/library/OriginWithdrawalLib.sol
```diff
@@ -0,0 +1,88 @@
...
+        IOETHVault oethVault = IOETHVault(MC.OETH_VAULT);
+
+        uint256 amountInOETH = woeth.redeem(amount, address(this), address(this));
+        oeth.approve(address(oethVault), amountInOETH);
+        (requestId,) = oethVault.requestWithdrawal(amountInOETH);
+
+        _addRequestId(requestId);
+
+        emit WOETHWithdrawalRequested(amount, requestId);
+    }
```
danoctavian, line 83: lets make a function requestWithdrawalOETH that also handles OETH and requestWithdrawalWOETH calls requestWithdrawalOETH after redeem so we have both options on the table

src/library/AsyncWithdrawLib.sol
danoctavian: we must solve for fees yes

src/library/AsyncWithdrawalLib.sol
```diff
@@ -0,0 +1,83 @@
...
+        IWithdrawalQueueManager queueManager = IWithdrawalQueueManager(queueManager_);
+        (, IWithdrawalQueueManager.WithdrawalRequest[] memory requests) =
+            queueManager.withdrawalRequestsForOwner(address(this));
+        for (uint256 i = 0; i < requests.length; i++) {
+            if (!requests[i].processed) {
+                // NOTE: needs to be fixed - assumes no slashing for now, as in reality eigenlayer slashing is not active yet
+                // also we do not account for the fees here
+                baseAssets += requests[i].amount * requests[i].redemptionRateAtRequestTime / 1e18;
+                // subtract fee
+                baseAssets -= queueManager.calculateFee(requests[i].amount, requests[i].feeAtRequestTime);
```
danoctavian: this is incorrect as it substracts the fee in the unit of measure of amount, which is ynETH or ynLSDe

while basAssets is already converted

apply the fee first and convert after

src/library/AsyncWithdrawalLib.sol
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    function _asyncWithdrawalBalanceYNAsset(address queueManager_) private view returns (uint256 baseAssets) {
+        IWithdrawalQueueManager queueManager = IWithdrawalQueueManager(queueManager_);
+        (, IWithdrawalQueueManager.WithdrawalRequest[] memory requests) =
+            queueManager.withdrawalRequestsForOwner(address(this));
+        for (uint256 i = 0; i < requests.length; i++) {
+            if (!requests[i].processed) {
+                // NOTE: needs to be fixed - assumes no slashing for now, as in reality eigenlayer slashing is not active yet
+                // also we do not account for the fees here
```
danoctavian: if fees are addressed delete the comment

test/mainnet/helpers/SetupWithdrawer.sol
```diff
@@ -0,0 +1,78 @@
...
+
+contract SetupWithdrawer is Test, Etches, MainnetActors, WithdrawerUtils {
+    function setup() public returns (Withdrawer vault) {
+        string memory name = "YieldNest Withdrawer";
+        string memory symbol = "ynWithdrawer";
+
+        Withdrawer vaultImplementation = new Withdrawer();
+
+        // Deploy the proxy
+        bytes memory initData =
```
danoctavian: can we simply call initialize separately, it's hard to read the parameters here and we lose type checking

directly on the contract

test/mainnet/helpers/SetupWithdrawer.sol
```diff
@@ -0,0 +1,78 @@
...
+            abi.encodeWithSelector(Withdrawer.initialize.selector, ADMIN, name, symbol, 18, true, true);
```
danoctavian: set alwaysComputeTotalAssets_ to false, that's how we want to use this in production

test/mainnet/helpers/SetupWithdrawer.sol
```diff
@@ -0,0 +1,78 @@
+// SPDX-License-Identifier: BSD Clause-3
+pragma solidity ^0.8.24;
+
+import {Test} from "lib/forge-std/src/Test.sol";
+import {Withdrawer} from "src/withdraws/Withdrawer.sol";
+import {TransparentUpgradeableProxy as TUProxy} from "src/Common.sol";
```
danoctavian: don't use the TUPProxy alias just call it explicitly TransparentUpgradeableProxy

test/mainnet/helpers/SetupWithdrawer.sol
danoctavian: do this for SetupWithdrawer.sol in the unit test directory

PR #82, opened 2025-01-24
sync eth-max-vault v0.2.0

dan13ram:


src/BaseVault.sol
```diff
@@ -472,9 +456,7 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
      * @return uint256 The equivalent amount in base denomination.
      */
     function _convertAssetToBase(address asset_, uint256 assets) internal view virtual returns (uint256) {
```
yield-farmer-1, line 458: this function is not used anywhere

src/BaseVault.sol
```diff
@@ -484,42 +466,31 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
      * @return uint256 The equivalent amount of assets.
      */
     function _convertBaseToAsset(address asset_, uint256 assets) internal view virtual returns (uint256) {
```
yield-farmer-1, line 468: this function is not used anywhere

src/BaseVault.sol
danoctavian: after moving the logic in VaultLib.sol the function is not called inside the vault anymore but the logic it runs is used inside the VaultLib

But when you inherit the vault, because it's internal, you can call it when you need it

same for _convertBaseToAsset

test/mainnet/withdrawer.spec.sol
```diff
@@ -0,0 +1,146 @@
...
+        assertApproxEqRel(vault.totalAssets(), totalAssets, 1e15, "Total assets should match");
+
+        vault.processAccounting();
+    }
+
+    function _claimWithdrawal(uint256 tokenId) internal {
+        address asset_ = MC.SLISBNB;
+        address stakeManager_ = MC.SLIS_BNB_STAKE_MANAGER;
+        ISlisBnbStakeManager stakeManager = ISlisBnbStakeManager(stakeManager_);
+        MockStakeHub stakeHub = MockStakeHub(MC.SLIS_BNB_STAKE_HUB);
```
danoctavian, line 126: Can we test this without a mock?

it's important here we test against the real thing

PR #85, opened 2025-01-24
Feat/move async eth max

MrDeadCe11:
moved changes to provider and async lib

### DIFF
--- a/src/module/Provider.sol
+++ b/src/module/Provider.sol
@@ -11,8 +11,12 @@ import {
     IFrxEthWethDualOracle,
     IynLSDe
 } from "src/interface/IProvider.sol";
-import {IERC4626} from "src/Common.sol";
+import {IERC4626, ERC20} from "src/Common.sol";
 import {MainnetContracts as MC} from "script/Contracts.sol";
+import {IWithdrawalQueueManager} from "src/interface/IWithdrawalQueueManager.sol";
+import {IWithdrawalQueue} from "src/interface/external/lido/IWithdrawalQueue.sol";
+import {IWithdrawer} from "src/interface/IWithdrawer.sol";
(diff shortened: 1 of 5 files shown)


script/Contracts.sol
```diff
@@ -46,4 +46,8 @@ library MainnetContracts {
     address public constant FRX_ETH_WETH_DUAL_ORACLE = 0x350a9841956D8B0212EAdF5E14a449CA85FAE1C0;
 
     address public constant WBTC = 0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599;
+
+    address public constant CURVE_LP_YNETH_YNLSDE_POOL = 0x1f59cC10c6360DA918B0235c98E58008452816EB;
```
dan13ram: Hey could we please remove these extra changes (i think they got moved over from the tokenized strategy branch)

script/Contracts.sol
test/mainnet/helpers/SetupWithdrawer.sol
test/unit/helpers/Etches.sol

src/module/Provider.sol
```diff
@@ -73,4 +77,78 @@ contract Provider is IProvider {
...
+            baseAssets += statuses[i].amountOfStETH;
+        }
+    }
+
+    function _asyncWithdrawalBalance(address asset) private view returns (uint256 baseAssets) {
+        if (asset == MC.WOETH) {
+            IOETHVault oethVault = IOETHVault(MC.OETH_VAULT);
+
+            // TODO: see if we should add the withdrawer contracts to Contracts.sol and use it here instead of msg.sender
+            uint256[] memory requestIds = IWithdrawer(msg.sender).getWOETHRequestIds();
```
dan13ram: This is probably the only place we need to think about a little bit. It's very specific to the OETH use case.. @danoctavian what's your opinion on this?

PR #87, opened 2025-01-25
add base withdraw + move async balance to withdrawer

dan13ram:


src/withdraws/BaseWithdrawer.sol
```diff
@@ -0,0 +1,57 @@
...
+    function _feeOnRaw(uint256) public pure override returns (uint256) {
+        return 0;
+    }
+
+    function _feeOnTotal(uint256) public pure override returns (uint256) {
+        return 0;
+    }
+
+    function computeTotalAssets() public view virtual override returns (uint256 totalBaseBalance) {
+        totalBaseBalance = AsyncWithdrawalLib.computeTotalAssets();
```
dan13ram: @danoctavian I'm not totally clear on how the lines for auditing are calculated. `AsyncWithdrawalLib` is pulled in here for `computeTotalAssets`. Does that mean all of AsyncWithdrawalLib is also audited?
Therefore should this override be in BaseWithdrawer or Withdrawer @danoctavian ?

I considered moving the new `computeTotalAssets` logic from the library to inside `BaseWithdrawer` itself. 
The size difference is 20,006 vs 19,788. Seems significant enough for it to be in the library itself.

Let me know what you think

danoctavian (approved):

PR #88, opened 2025-01-26
added fork tests for LP connector and strategy

dan13ram:


script/Contracts.sol
```diff
@@ -24,7 +24,7 @@ library MainnetContracts {
     address public constant PROXY_ADMIN = 0xA02A8DC24171aC161cCb74Ef02C28e3cA2204783;
 
     address public constant PROVIDER = address(123456789); // TODO: Update with deployed Provider
-    address public constant BUFFER = address(987654321); // TODO: Update with deployed buffer
+    address public constant BUFFER = address(0x45c3B59d53e2e148Aaa6a857521059676D5c0489);
```
dan13ram: @danoctavian I added the BUFFER strategy address here since you said this will be the one used? Is that okay ?

src/module/Provider.sol
```diff
@@ -72,6 +75,19 @@ contract Provider is IProvider {
             return IsfrxETH(MC.SFRXETH).pricePerShare() * frxETHPriceInETH / 1e18;
         }
 
+        if (asset == MC.CURVE_LP_YNETH_YNLSDE_STRATEGY) {
+            (int256 lpRate, uint256 updatedAt) = ICurveLpConnector(MC.CURVE_LP_YNETH_YNLSDE_CONNECTOR).rate();
+            if (lpRate < 0) {
+                revert RateIsNegative();
+            }
+            // solhint-disable-next-line not-rely-on-time
+            if (updatedAt < block.timestamp - 24 hours) {
```
dan13ram: also this check, do you think its important to have?

test/mainnet/buffer.spec.sol
```diff
@@ -11,8 +11,9 @@ import {AssertUtils} from "test/utils/AssertUtils.sol";
 import {MockERC4626} from "test/mainnet/mocks/MockERC4626.sol";
 import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import {MockProvider} from "test/unit/mocks/MockProvider.sol";
+import {BaseRules} from "script/rules/BaseRules.sol";
 
-contract VaultMainnetInvariantsTest is Test, AssertUtils, MainnetActors {
+contract VaultMainnetInvariantsTest is Test, AssertUtils, MainnetActors, BaseRules {
```
dan13ram, line 16: the changes to the rest of the tests are refactor related changes for the rules.

test/mainnet/mocks/MockConnector.sol
```diff
@@ -0,0 +1,104 @@
...
+    ICurvePool public immutable CURVE_POOL;
+    IStrategy public immutable STRATEGY;
+
+    IERC20 public immutable ASSET_A;
+    IERC20 public immutable ASSET_B;
+
+    int256 public storedRate;
+    uint256 public timestamp;
+
+    constructor(address _vault, address _strategy, address _assetA, address _assetB) {
```
dan13ram, line 35: Copied over the code of Connector and modified some things for easy mocking.

script/Contracts.sol
danoctavian: can we define a named constant separately?  with EULER_and_what_it_is

src/module/Provider.sol
danoctavian: no it's fine let's leave it out currently it doesn't get stale

good point though

test/mainnet/tokenizedlpstrategy.t.sol
```diff
@@ -0,0 +1,271 @@
...
+
+        uint256 previewedShares = vault.previewDeposit(assets);
+        assertApproxEqRel(previewedShares, shares, 1e15, "Previewed shares should equal the converted shares");
+
+        // Test the previewMint function
+        uint256 previewedAssets = vault.previewMint(shares);
+        assertApproxEqRel(previewedAssets, assets, 1e15, "Previewed assets should equal the original assets");
+
+        {
+            deal(MC.WETH, address(this), assets);
```
danoctavian: don't use address(this) , designate a `depositor` or `bob` or `alice` as address(this) is easy to get it mixed together between different interactions

src/module/Provider.sol
```diff
@@ -72,6 +75,19 @@ contract Provider is IProvider {
...
+                revert RateIsStale();
+            }
+
+            return uint256(lpRate);
```
danoctavian: the rate required here is not the rate of the LP but the rate of the strategy

so far you have the rate of the LP, you need to combine the 2 rates, the rate of the strategy with the rate of the lp

see here how this does it:

https://github.com/yieldnest/curve-lp-ynethx/pull/1/files#diff-f51577955a762ffe60e506c06158b6b0cb53818c1215bc50d5e1374ce39e7fcbR46

dan13ram: @danoctavian I know we discussed that. But this is actually already calculated 

https://github.com/yieldnest/curve-lp-ynethx/blob/04ff958bc371375c8da74d771723024578d05aec/src/test/Connector.t.sol#L32

If you see here the `YvCurveLpOracle` oracle is used in the deployment of Connector.sol 

https://github.com/yieldnest/curve-lp-ynethx/blob/04ff958bc371375c8da74d771723024578d05aec/src/periphery/YvCurveLpOracle.sol

And the convertToShares is already done in this oracle.

test/mainnet/tokenizedlpstrategy.t.sol
dan13ram: fixed in [742159e](https://github.com/yieldnest/yieldnest-vault/pull/88/commits/742159e4f34ba39e9ca27f30c84a4de4edfcf847)

src/module/Provider.sol
dan13ram: removed in [742159e](https://github.com/yieldnest/yieldnest-vault/pull/88/commits/742159e4f34ba39e9ca27f30c84a4de4edfcf847)

script/Contracts.sol
dan13ram: updated in [742159e](https://github.com/yieldnest/yieldnest-vault/pull/88/commits/742159e4f34ba39e9ca27f30c84a4de4edfcf847)

src/module/Provider.sol
dan13ram: I renamed the variable to `strategyRate` for clarity.

danoctavian: you are correct

PR #91, opened 2025-02-03
added ynETHx configurer

dan13ram:


test/mainnet/configurer.spec.sol
```diff
@@ -0,0 +1,109 @@
...
+import {MainnetActors} from "script/Actors.sol";
+import {MainnetContracts as MC} from "script/Contracts.sol";
+import {YnETHxVault} from "src/YnETHxVault.sol";
+
+import {YnETHxConfigurer} from "src/configures/YnETHxConfigurer.sol";
+
+import {SetupWithdrawer} from "test/mainnet/helpers/SetupWithdrawer.sol";
+
+contract VaultConfigureUpgradeTest is Test, MainnetActors {
+    function test_configure() public {
```
dan13ram, line 23: Added this test to check the gas costs for `YnETHxConfigurer.configure()` call using 
```bash
FOUNDRY_PROFILE=mainnet forge test --gas-report --match-test test_configure
```

The result is that `YnETHxConfigurer.configure()` costs 4656783 which is well under 30M gas.

script/rules/SafeRules.sol
```diff
@@ -0,0 +1,34 @@
...
+    {
+        setProcessorRule(vault_, contractAddress, funcSig, rule, false);
+    }
+
+    function setProcessorRule(
+        IVault vault_,
+        address contractAddress,
+        bytes4 funcSig,
+        IVault.FunctionRule memory rule,
+        bool force
```
dan13ram: added this to ensure that we don't write over already set rules in our scripts.

src/module/Provider.sol
```diff
@@ -14,15 +14,26 @@ import {
...
 
+interface IBaseStrategy {
+    function STRATEGY_VERSION() external view returns (string memory);
+}
+
 contract Provider is IProvider {
     error UnsupportedAsset(address asset);
     error RateIsNegative();
 
+    function isVault(address asset) public view returns (bool) {
```
dan13ram, line 29: This extra `isVault` is needed since the withdrawer is not deployed and we are using that as one of the strategies. I'm not particularly happy about this approach.

The other option is to deploy the withdrawer and add `MC.WITHDRAWER` address to the Contracts library.

src/configures/YnETHxConfigurer.sol
```diff
@@ -0,0 +1,164 @@
...
+import {BaseRules} from "script/rules/BaseRules.sol";
+import {ConnectorRules} from "script/rules/ConnectorRules.sol";
+import {YieldNestRules} from "script/rules/YieldNestRules.sol";
+import {BaseRoles} from "script/roles/BaseRoles.sol";
+
+contract YnETHxConfigurer is MainnetActors {
+    error NotAdmin();
+    error InvalidVaultVersion();
+
+    function configure(address provider, address withdrawer) external {
```
danoctavian: what's the gas consumption of running this? as reported by --gas-report

script/roles/BaseRoles.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: BSD Clause-3
+pragma solidity ^0.8.24;
+
+import {BaseVault} from "src/BaseVault.sol";
+import {IActors} from "script/Actors.sol";
+
+library BaseRoles {
```
danoctavian, line 8: good use of reusable code

test/mainnet/helpers/SetupVault.sol
```diff
@@ -24,7 +28,7 @@ contract SetupVault is Test, MainnetActors, Etches, BaseRules {
 
         bytes4 selector = bytes4(keccak256("upgradeAndCall(address,address,bytes)"));
 
-        bytes memory initData = abi.encodeWithSelector(ynETHxVault.initializeV2.selector, 18, 0);
+        bytes memory initData = abi.encodeWithSelector(YnETHxVault.initializeV2.selector, 18, 0);
```
danoctavian, line 31: i don't think we'll be able to use this script as something that sends to the timelock we'd just load up the data manually

test/mainnet/invariants.spec.sol
```diff
@@ -21,22 +20,22 @@ contract VaultMainnetInvariantsTest is Test, AssertUtils, MainnetActors {
 
     function totalSupplyInvariant(uint256 supply) public view {
         uint256 finalVaultTotalSupply = vault.totalSupply();
-        assertEqThreshold(
-            supply, finalVaultTotalSupply, 3, "Vault totalSupply should be original totalSupply plus additional"
+        assertApproxEqRel(
+            supply, finalVaultTotalSupply, 1e14, "Vault totalSupply should be original totalSupply plus additional"
```
danoctavian, line 27: why is this done what changed?

test/mainnet/invariants.spec.sol
```diff
@@ -21,22 +20,22 @@ contract VaultMainnetInvariantsTest is Test, AssertUtils, MainnetActors {
...
         );
     }
 
     function totalAssetsInvariant(uint256 assets) public view {
         uint256 finalVaultTotalAssets = vault.totalAssets();
-        assertEqThreshold(
-            assets, finalVaultTotalAssets, 3, "Vault totalAssets should be original totalAssets plus additional"
+        assertApproxEqRel(
```
danoctavian, line 33: same here if these changes are added they should be justified

src/configures/YnETHxConfigurer.sol
danoctavian: concerned about whether we need to make this an iterative run if gas is ridiculously high

which will complicate it so it's better to measure asap

test/mainnet/invariants.spec.sol
dan13ram: With the addition of more assets and particularly the lp pool strategy, the variations are a little above the "3" threshold that was there earlier. I changed it to use `assertApproxEqRel` with 0.01 % threshold to have a more deterministic test

script/rules/BaseRules.sol
```diff
@@ -2,16 +2,25 @@
 pragma solidity ^0.8.24;
 
 import {IVault, IValidator} from "src/interface/IVault.sol";
+import {SafeRules} from "./SafeRules.sol";
 
-contract BaseRules {
+library BaseRules {
```
danoctavian, line 7: as discussed in DMs we want to have a common trunk for these libraries that simply generate the rules in data form.

A common format here may be  as an example

`struct Rule {
address target,
bytes4 funcSig,
IVault.FunctionRule memory rule
}`

as a data representation of the rule

Here's a prototype script that will use these libs (no need for compatibility with this, please choose the best naming structure you see fit):

https://github.com/yieldnest/yieldnest-vault/blob/b501a7e9f2be038a399a37f6c0becb6b17461e00/script/rules/GenerateAsBNBRulesTxData.s.sol

The other use case is as you have it here in the Configurer or a fresh deploy script where the rules are crrated and also set.

https://github.com/yieldnest/yieldnest-vault/blob/bnb-max-vault/script/DeployMaxVault.s.sol#L133

src/configures/YnETHxConfigurer.sol
dan13ram:
```bash
FOUNDRY_PROFILE=mainnet forge test --gas-report --match-test test_configure
```

Screenshot 2025-02-05 at 2 15 04 PM

danoctavian: good thanks!

danoctavian:
we also need to be able to deploy the Withdrawer + rules

We can do that script in this PR to see if the rules at work, and fill in the gaps there too

PR #96, opened 2025-02-13
Fix/provider is vault

danoctavian:


src/module/Provider.sol
```diff
@@ -26,10 +28,10 @@ contract Provider is IProvider {
     error UnsupportedAsset(address asset);
     error RateIsNegative();
 
-    function isVault(address asset) public view returns (bool) {
+    function isETHStrategyVault(address asset) public view returns (bool) {
         try IBaseStrategy(asset).STRATEGY_VERSION() returns (string memory version) {
-            // TODO: this is not ideal, as it's not forward compatible with future vault versions ...
-            return keccak256(bytes(version)) == keccak256(bytes("0.1.0"));
+            address vaultAsset = IVault(asset).asset();
+            return keccak256(bytes(version)) == keccak256(bytes("0.1.0")) && vaultAsset == MC.WETH;
```
danoctavian, line 34: if the base asset is not WETH the conversion is wrong (eg. suppose the vault is denominated in WSTETH (doesn't exist but in theory it can)

PR #101, opened 2025-02-14
additional tests + deploy configurer

### LINKED ISSUE #98: Additional verifications for configurer.spec.sol
for the configurer.spec.sol 

add some additional verifications 
moves the upgrade execution to setup()

adds a few tests to :

deposit  ynETH, ynLSDe, wETH and assert totalAssets() is correct

donate the 10 whitelisted assets to the vault and see that  totalAssets() is correct, (post calling processAccounting(

Allocate to buffer and withdraw once and assert

Move a few assets to withdrawer call processAccounting there, call it on the ynETHx vault assert totalAssets() is correct

Additionally if we can edit the Withdrawer script to also deploy the provider and configurer so we have them all in One go

MrDeadCe11:
adding additional configurer tests and more deployment scripts

closes #98


test/mainnet/configurer.spec.sol
```diff
@@ -149,4 +158,129 @@ contract VaultConfigureUpgradeTest is Test, MainnetActors {
...
+
+        whitelistedAssets[6] = MC.WSTETH;
+        vm.label(MC.WSTETH, "WSTETH");
+        deal(MC.WSTETH, address(this), donationAmount);
+        
+        whitelistedAssets[7] = MC.OETH;
+        vm.label(MC.OETH, "OETH");
+        // {
+        //  //TODO: figure out how to get OETH this reverts
+        //     IERC20(MC.WETH).approve(MC.OETH, donationAmount);
```
danoctavian: needs to fix this

dan13ram: fixed in [900f7b4](https://github.com/yieldnest/yieldnest-vault/pull/101/commits/900f7b459f7c0d067afd5ee0e90f2fc1ab85cd78)

script/rules/YieldNestRules.sol
```diff
@@ -32,17 +32,17 @@ library YieldNestRules {
     {
         bytes4 funcSig = bytes4(keccak256("deposit(address,uint256,address)"));
 
-        IVault.ParamRule[] memory paramRules = new IVault.ParamRule[](2);
+        IVault.ParamRule[] memory paramRules = new IVault.ParamRule[](3);
```
danoctavian, line 35: so there was a bug here where rule was incomplete

danoctavian (approved):

PR #107, opened 2025-02-16
deployment of ynETHx v1

danoctavian:


script/VerifyMaxVault.s.sol
```diff
@@ -20,15 +20,21 @@ contract VerifyMaxVault is BaseScript, Test {
     }
 
     function run() public {
-        _loadDeployment();
         _setup();
-
+        _loadDeployment();
+        assertNotEq(msg.sender, deployer, "msg.sender should not be deploye as this is a verifier script.");
```
danoctavian, line 25: this was added so that it doesn't accidentally inject the default foundry msg.sender in there which is what gets added normally

and it parses out the one from the json file.

script/VerifyMaxVault.s.sol
```diff
@@ -96,14 +109,35 @@ contract VerifyMaxVault is BaseScript, Test {
         RolesVerification.verifyTemporaryRoles(vault, deployer);
         RolesVerification.verifyTemporaryRoles(withdrawer, deployer);
 
+        // FIXME: TODO: reenable this once viewer is deployed
         // verify viewer
-        VaultVerification.verifyViewer(viewer, vault);
-        assertTrue(
-            MaxVaultViewer(address(viewer)).isUnderlyingAsset(contracts.WETH()), "WETH should be an underlying asset"
-        );
+        // VaultVerification.verifyViewer(viewer, vault);
```
danoctavian, line 114: this needs to be reenabled once we deploy viewer which will be done after upgrade.

script/verification/RolesVerification.sol
```diff
@@ -21,7 +21,7 @@ library RolesVerification {
         Vm vm = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
 
         bool hasRole = control.hasRole(role, account);
-        console.log(hasRole ? "\u2705" : "\u274C", message, account);
+        console.log(hasRole == expected ? "\u2705" : "\u274C", message, account);
```
danoctavian, line 24: this was printing ❌  when it was correct because assertion here is that the role is not with this address

test/mainnet/configurer.spec.sol
```diff
@@ -31,93 +31,9 @@ contract VaultConfigureUpgradeTest is TestHelper, MainnetActors {
 
     function setUp() public {
         vault = Vault(payable(MC.YNETHX));
-        _initVault(vault);
```
danoctavian, line 34: delete all upgrade related things

test/mainnet/configurer.spec.sol
```diff
@@ -233,75 +149,40 @@ contract VaultConfigureUpgradeTest is TestHelper, MainnetActors {
...
-        if (assets[i] == MC.STETH) {
-            vm.assume(donationAmount < 1000 ether);
-        } else {
-            vm.assume(donationAmount < 1e5 ether);
+        for (uint256 i = 0; i < assets.length; i++) {
+            _test_donate_single_asset(assets[i], donationAmount);
         }
-
-        // empty vault so we can test each asset in isolation
-        _emptyVault();
```
danoctavian, line 267: this test had some errors such as emptying the vault mid-test which creates a very particular testing environment (eg . totalShares are 1 million, totalAssets is 0) which doesn't reflect well what's being tested

test/mainnet/helpers/TestHelper.sol
```diff
@@ -19,10 +19,24 @@ contract TestHelper is Test {
...
         _vault = vault_;
     }
+    /// @notice Deals assets to an account by converting ETH to the desired asset
+    /// @param asset The asset to deal
+    /// @param account The account to receive the asset
+    /// @param amount The amount of ETH to convert
+    /// @return The actual amount of asset received after conversion
 
     function dealAsset(address asset, address account, uint256 amount) internal returns (uint256) {
+        if (asset == MC.WETH) {
```
danoctavian, line 29: small improvement here

PR #108, opened 2025-02-17
added script for deploying viewer

dan13ram:


script/DeployViewer.s.sol
```diff
@@ -0,0 +1,87 @@
...
+        // setup underlying assets
+        address[] memory underlyingAssets = new address[](5);
+        underlyingAssets[0] = MC.WETH;
+        underlyingAssets[1] = MC.STETH;
+        underlyingAssets[2] = MC.WSTETH;
+        underlyingAssets[3] = MC.OETH;
+        underlyingAssets[4] = MC.WOETH;
+
+        // NOTE: the following call will fail is ynETHx vault is not upgraded
+        viewer.addUnderlyingAssets(underlyingAssets);
```
dan13ram, line 75: @danoctavian this call will fail now since the vault isn't upgraded yet. 
Either we comment this out and call addUnderlyingAssets as the multisig post upgrade 
OR we upgrade the vault first then deploy the viewer fresh.

danoctavian: it's fine to do it after vault upgrade thanks for the note

PR #109, opened 2025-02-18
Test/extra process accounting tests

danoctavian:


test/mainnet/invariants.spec.sol
```diff
@@ -84,19 +83,19 @@ contract VaultMainnetInvariantsTest is Test, MainnetActors {
 
         // Test the convertToAssets function
         uint256 convertedAssets = vault.convertToAssets(shares);
-        assertApproxEqRel(convertedAssets, assets, 1e14, "Converted assets should equal the original assets");
+        assertApproxEqAbs(convertedAssets, assets, 3, "Converted assets should equal the original assets");
```
danoctavian, line 86: relative error used was very high because at low wei amounts eg. 1000, 2 wei of error is 0.2%. 

but that error does not scale with size of input.

test/mainnet/invariants.spec.sol
```diff
@@ -218,22 +217,26 @@ contract VaultMainnetInvariantsTest is Test, MainnetActors {
...
 
         // Test the previewRedeem function
         uint256 previewedRedeemAssets = vault.previewRedeem(shares);
-        assertApproxEqRel(
-            previewedRedeemAssets, assets, 2e15, "Previewed redeem assets should equal the original assets"
+        // TODO: see why precision is not high here.
+        assertApproxEqAbs(
+            previewedRedeemAssets,
+            assets - FeeMath.feeOnTotal(assets, vault.baseWithdrawalFee()),
+            100000,
```
danoctavian, line 238: still questionable why this is so high but notably more acceptable than a 0.2% error tolerance which is absurdly high

test/mainnet/invariants.spec.sol
```diff
@@ -259,7 +262,7 @@ contract VaultMainnetInvariantsTest is Test, MainnetActors {
     }
 
     function test_Vault_4626Invariants_withdraw(uint256 assets) public {
-        if (assets < 100_000) return;
+        if (assets < 100_000_000) return;
```
danoctavian, line 270: large relative errors appear when amounts are low, this may not be necessary

PR #110, opened 2025-02-18
reduced thresholds in ynethx tests

dan13ram:


test/mainnet/withdrawer.spec.sol
```diff
@@ -16,12 +15,13 @@ import {IProvider} from "src/interface/IProvider.sol";
...
 import {IOETHVault} from "src/interface/external/origin/IOETHVault.sol";
+import {TestHelper} from "test/mainnet/helpers/TestHelper.sol";
 
-contract WithdrawerMainnetTest is Test, MainnetActors {
+contract WithdrawerMainnetTest is TestHelper, MainnetActors {
     using Math for uint256;
 
     Withdrawer public vault;
-    uint256 public constant INITIAL_BALANCE = 100 ether;
+    uint256 public constant INITIAL_BALANCE = 1e5 ether;
```
danoctavian: if we have things like these let's write them

10_000 ether

for better readability.

i mean specifically for these cases there is XYZ ether.

danoctavian (approved):

PR #111, opened 2025-02-19
fixed empty vault in configurer spec test

dan13ram:


danoctavian:
Already fixed in deploy branch, no problem, will close this as a result

PR #116, opened 2025-03-08
Fix/base asset decimals issue

danoctavian:
# Issues

### issue 1 - first mint of a vault with 18 decimals that has a base asset with less than 18 decimals

This first mint mints an amount equal to the deposited amount

Example if you deposit 1 USDC in a ynUSDx vault with 18 decimals, it mints 1e6 even though the vault has 18 decimals, which starts the vault with a very low amount of shares (inesthetic, non-user friendly, may cause confusion).

### issue 2 - the rate provided by the vault in a multi-asset setup is losing precision if the base asset has 6 decimals and others have 18, and the rate is in base asset

eg. a Base vault with USDC as a base asset, and with USDE, DAI, SUSDE in the basket (18 decimal USD assets).

### issue 3 - the convertToShares calculation rounds down when a baseAsset with 6 decimals is used in a vault his 18 decimals

This can be a problem when burning shares and cause burning 2 few shares which can open up to a draining attack.

https://github.com/yieldnest/yieldnest-vault/blob/3f32414d2e60d19d86d4841928b0a3bfccf0e29b/src/library/VaultLib.sol#L246

### DIFF
--- a/src/library/VaultLib.sol
+++ b/src/library/VaultLib.sol
@@ -218,7 +218,17 @@ library VaultLib {
     {
         uint256 totalAssets = IVault(address(this)).totalAssets();
         uint256 totalSupply = getERC20Storage().totalSupply;
-        baseAssets = shares.mulDiv(totalAssets + 1, totalSupply + 1, rounding);
+
+        // Handle the case when totalAssets or totalSupply is zero
+        if (totalAssets == 0 || totalSupply == 0) {
+            IVault.VaultStorage storage vaultStorage = getVaultStorage();
+            address baseAsset = getAssetStorage().list[0];
+            baseAssets =
+                shares.mulDiv(1, 10 ** (vaultStorage.decimals - IERC20Metadata(baseAsset).decimals()), rounding);
+        } else {
+            baseAssets = shares.mulDiv(totalAssets + 1, totalSupply + 1, rounding);
+        }
+
         assets = convertBaseToAsset(asset_, baseAssets);
     }
 
--- a/src/library/VaultLib.sol
+++ b/src/library/VaultLib.sol
@@ -219,7 +219,7 @@ library VaultLib {
             uint256 baseAssetsOffset = 10 ** (vaultStorage.decimals - IERC20Metadata(baseAsset).decimals());
             baseAssets = shares.mulDiv(1, baseAssetsOffset, rounding);
         } else {
-            baseAssets = shares.mulDiv(totalAssets + 1, totalSupply + 1, rounding);
+            baseAssets = shares.mulDiv(totalAssets, totalSupply, rounding);
         }
 
         assets = convertBaseToAsset(asset_, baseAssets);
--- a/src/library/VaultLib.sol
+++ b/src/library/VaultLib.sol
@@ -244,8 +244,6 @@ library VaultLib {
 
         // Handle the case when totalAssets or totalSupply is zero
         if (totalAssets == 0 || totalSupply == 0) {
-            IVault.VaultStorage storage vaultStorage = getVaultStorage();
-
             return (baseAssetsWithOffset, baseAssets);
         }
 
--- a/src/library/VaultLib.sol
+++ b/src/library/VaultLib.sol
@@ -3,7 +3,7 @@ pragma solidity ^0.8.24;
 
 import {IVault} from "src/interface/IVault.sol";
 import {IProvider} from "src/interface/IProvider.sol";
-import {Math, IERC20} from "src/Common.sol";
+import {Math, IERC20, IERC20Metadata} from "src/Common.sol";
 import {Guard} from "src/module/Guard.sol";
 
 library VaultLib {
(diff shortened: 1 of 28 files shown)


src/library/VaultLib.sol
```diff
@@ -218,7 +223,15 @@ library VaultLib {
         uint256 totalAssets = IVault(address(this)).totalAssets();
         uint256 totalSupply = getERC20Storage().totalSupply;
         uint256 baseAssets = convertAssetToBase(asset_, assets);
-        uint256 shares = baseAssets.mulDiv(totalSupply + 1, totalAssets + 1, rounding);
+
+        // Handle the case when totalAssets or totalSupply is zero
+        if (totalAssets == 0 || totalSupply == 0) {
+            IVault.VaultStorage storage vaultStorage = getVaultStorage();
+            address baseAsset = getAssetStorage().list[0];
+            return (baseAssets * 10 ** (vaultStorage.decimals - IERC20Metadata(baseAsset).decimals()), baseAssets);
```
yield-farmer-1: this basically assumes that `vaultStorage.decimals` should always be `>=18` otherwise it can revert. Is it ok if we hardcode `vaultStorage.decimals` to `18` to avoid any confusion. Maybe we should hardcode our `BaseVault` to only support `18` decimals to avoid any issues

test/unit/vault/base6decimals/deposit.t.sol
```diff
@@ -0,0 +1,97 @@
...
+            IERC20(MC.USDC).balanceOf(alice),
+            INITIAL_BALANCE - depositAmount,
+            "Alice's balance did not decrease correctly"
+        );
+
+        // Check that Alice received the correct amount of shares
+        assertEq(vault.balanceOf(alice), sharesMinted, "Alice did not receive the correct amount of shares");
+
+        // Check that shares minted is depositAmount * 1e12 (converting from 6 to 18 decimals)
+        assertEq(sharesMinted, depositAmount * 1e12, "Incorrect number of shares minted");
```
yield-farmer-1, line 92: can we also add a check that `vault.convertToAssets` results in expected value

test/unit/vault/base6decimals/SetupBase6DecimalsVault.sol
```diff
@@ -0,0 +1,79 @@
...
+        setApprovalRule(vault, address(vault), MC.BUFFER);
+        setApprovalRule(vault, MC.WETH, MC.BUFFER);
+
+        vault.setBuffer(MC.BUFFER);
+
+        // Set rates in provider
+        MockProvider(MC.PROVIDER).setRate(MC.USDC, 1e6); // 1 USD USDC
+        MockProvider(MC.PROVIDER).setRate(MC.WBTC, 100_000e6); // 100k USD bitcoin
+        MockProvider(MC.PROVIDER).setRate(MC.STETH, 10_000e6); // 10k USD steth
+
```
yield-farmer-1: Is there any general guideline on how should we determine rate and decimals in `Provider`? The logic of determining rate in terms of baseAsset will not work in general case.

For ex: consider Vault with `wBTC` as baseAsset. In that case, how will we determine rate of `USDC`(assuming vault supports it) in terms of `wBTC`? it would be less than `1` which won't work. So, in general this `Provider` model will not work in case where `1 unit(wei)` of baseAsset is more valuable than `1 unit` of other supported assets. For ex: 1 sats of wbtc is more valuable than 1 wei of usdc and hence it won't work in example I provide.

src/library/VaultLib.sol
danoctavian: i think we shouldn't hardcode it

test/unit/vault/base6decimals/SetupBase6DecimalsVault.sol
danoctavian: we so far just made the rate have same decimals as base Asset

but we lose precision a lot if that asset has a low decimal count

src/library/VaultLib.sol
```diff
@@ -91,6 +91,16 @@ library VaultLib {
             revert IVault.InvalidNativeAssetDecimals(decimals_);
         }
 
+        // Check that asset decimals are not greater than 18
+        if (decimals_ > 18) {
+            revert IVault.InvalidAssetDecimals(decimals_);
+        }
+
+        // Check if trying to add the primary asset again
+        if (index > 0 && asset_ == assetStorage.list[0]) {
```
danoctavian: this is a bug in addAsset you can add the baseAsset again

src/library/VaultLib.sol
```diff
@@ -135,6 +145,13 @@ library VaultLib {
         assetStorage.list[index] = assetStorage.list[assetStorage.list.length - 1];
         assetStorage.list.pop();
         delete assetStorage.assets[asset_];
+
+        // Update the index for the asset that was moved to the deleted position
+        if (index < assetStorage.list.length) {
```
danoctavian: another bug is the index doesn't get set correctly

src/library/VaultLib.sol
```diff
@@ -91,6 +91,11 @@ library VaultLib {
             revert IVault.InvalidNativeAssetDecimals(decimals_);
         }
 
+        // Vault does not support assets with more than 18 decimals
+        if (decimals_ > 18) {
```
danoctavian, line 95: Should this be `decimals_ > vaultStorage.decimals` as in no asset has more decimals than the vault instead?

That does imply if you have a vault with 8 decimals like ynBfBTCk you can't add an asset that has 18 decimals like BTCB

src/library/VaultLib.sol
```diff
@@ -199,7 +206,17 @@ library VaultLib {
     {
         uint256 totalAssets = IVault(address(this)).totalAssets();
         uint256 totalSupply = getERC20Storage().totalSupply;
-        baseAssets = shares.mulDiv(totalAssets + 1, totalSupply + 1, rounding);
+
+        // Handle the case when totalAssets or totalSupply is zero
+        if (totalAssets == 0 || totalSupply == 0) {
```
danoctavian, line 211: this is still potentially problematic.

What if these conditions are hit after the initial deposit?

It implies we mint a 1:1, is that still accurate?

PR #117, opened 2025-03-14
added strategy tests for conversion rates

dan13ram:


src/strategy/BaseStrategy.sol
```diff
@@ -104,7 +104,7 @@ abstract contract BaseStrategy is BaseVault, IBaseStrategy {
      * @return maxAssets The maximum amount of assets.
      */
     function _maxWithdrawAsset(address asset_, address owner) internal view virtual returns (uint256 maxAssets) {
-        if (paused() || !_getAssetStorage().assets[asset_].active) {
+        if (paused() || !_getBaseStrategyStorage().isAssetWithdrawable[asset_]) {
```
danoctavian, line 107: let's move this fix in a separate clean Pull request please with tests that are unit and integration that verify cover this

danoctavian:
I eliminated all assertApproxAbs clauses and tests still pass so these tests don't capture the precision issue

I think it may have to do with the rates of each asset all being set the same

src/strategy/BaseStrategy.sol
danoctavian: can we please revert this change here

PR #120, opened 2025-03-17
fix addAsset and deleteAsset

danoctavian:
Issue description:

Issue 1:  addAsset Can add a duplicate of the base asset (asset at index 0)
Issue 2: deleteAsset does not update the index in storage of the popped asset at the end of the list


dan13ram (approved):
Looks good. Should we update the vault version ?

PR #123, opened 2025-04-20
Fix/base asset decimals issue wrapper fix

danoctavian:
Solution:

Introduce the concept of defaultAssetIndex_

Therefore the base asset and the default Asset can be 2 different assets.

The Base Asset is the denomination Asset and is the asset at positon 0 in the assets array.

The Default Asset is the vault.asset() that is the default for the deposit() redeem(), withdraw() and mint().

### Details

As the vault was originally conceived, we had the notion of Base Asset. That means totalAssets evaluation and internal accounting converts the value to the base Asset.

The Base Asset was also the default asset for the deposit(), withdraw(), redeem(), mint() and auxilliary functions that follow the ERC4626 interface. 

This works well for ynBNBx and ynETHx as WBNB and WETH respectively.

However in the case of ynUSDx where an asset like USDC or USDT (6 decimals) are chosen as the base Asset, using USDC  as base asset becomes problematic as the internal accounting does all conversions to an asset with less precision than others (USDE, Sky Dollar which have 18 decimals), which causes high totalAssets computation imprecision, and implicitly rounds down due to truncation to 6 decimals.

The proposal is simple:

Separate the ERC4626 *default asset* from the base asset. 

Therefore one can use a base asset such as a WUSDC which has 18 decimals (by wrapping the raw USDC) and still expose an inteface that has USDC as `.asset().

This works well during testing, and implies relatively minimal changes to the BaseVault.

### DIFF
--- a/test/unit/strategy/base6decimals/withdraw.t.sol
+++ b/test/unit/strategy/base6decimals/withdraw.t.sol
@@ -1,0 +1,84 @@
+// SPDX-License-Identifier: BSD Clause-3
+pragma solidity ^0.8.24;
+
+import {Test} from "lib/forge-std/src/Test.sol";
+import {Vault} from "src/Vault.sol";
+import {TransparentUpgradeableProxy} from "src/Common.sol";
+import {MainnetContracts as MC} from "script/Contracts.sol";
+import {Etches} from "test/unit/helpers/Etches.sol";
+import {WETH9} from "test/unit/mocks/MockWETH.sol";
+import {SetupVault} from "test/unit/helpers/SetupVault.sol";
+import {MainnetActors} from "script/Actors.sol";
+import {MockSTETH} from "test/unit/mocks/MockST_ETH.sol";
+import {IVault} from "src/interface/IVault.sol";
+import {MockERC20} from "test/unit/mocks/MockERC20.sol";
+import {IERC4626} from "src/Common.sol";
+import {Provider} from "src/module/Provider.sol";
+import {IERC20} from "src/Common.sol";
+import {IProvider} from "src/interface/IProvider.sol";
+import {XReferralAdapter} from "src/utils/XReferralAdapter.sol";
+import {SetupBase6DecimalsVault} from "test/unit/vault/base6decimals/SetupBase6DecimalsVault.sol";
+import {BaseRules} from "script/rules/BaseRules.sol";
+import {SafeRules} from "script/rules/SafeRules.sol";
+import {PublicViewsVault} from "test/unit/helpers/PublicViewsVault.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+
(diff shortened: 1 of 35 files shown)


test/unit/vault/views.t.sol
```diff
@@ -196,6 +206,14 @@ contract VaultViewsUnitTest is Test, Etches {
         _testConvertToSharesForAsset(MC.WETH, assets, depositedAssets, rewards, 1e18);
     }
 
+    function test_Vault_convertToSharesForAsset_Concrete() public {
```
yield-farmer-1: can we use fuzzed input here?

test/unit/strategy/base6decimals/withdraw.t.sol
```diff
@@ -0,0 +1,84 @@
...
+import {IProvider} from "src/interface/IProvider.sol";
+import {XReferralAdapter} from "src/utils/XReferralAdapter.sol";
+import {SetupBase6DecimalsVault} from "test/unit/vault/base6decimals/SetupBase6DecimalsVault.sol";
+import {BaseRules} from "script/rules/BaseRules.sol";
+import {SafeRules} from "script/rules/SafeRules.sol";
+import {PublicViewsVault} from "test/unit/helpers/PublicViewsVault.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+
+contract BaseStrategy6DecimalsBaseDepositUnitTest is Test, MainnetActors, Etches {
+// function test_Vault_withdrawUSDE_afterUSDCDeposit() public {
```
yield-farmer-1: can we uncomment this test?

test/unit/vault/views.t.sol
danoctavian: deleted the test, it was just copy pasta from test_Vault_convertToSharesForAsset_WETH which is already fuzzed, for debugging

test/unit/strategy/base6decimals/withdraw.t.sol
danoctavian: yeah i'll beef up this whole suite thanks

xhad (commented):
This PR is opened against [eth-max-vault-release-candidate](https://github.com/yieldnest/yieldnest-vault/tree/eth-max-vault-release-candidate) but it's going to be the `ynUSDx` correct? 

Has the repo management changed from having 4 branches: eth-max-vault, btc-max-vault, usd-max-vault and bnb-max-vault?

danoctavian:
@xhad  no, we're still operating with `eth-max-vault, btc-max-vault, usd-max-vault and bnb-max-vault` 

But i introduced this release-candidate specific to eth-max-vault because there's multiple bug fixes in there that were worked on independently, so it made sense to have a release-candidate specific to `eth-max-vault` branch

We followed the procedure that you proposed initially, and files get copied over to the other branches

ynUSDx is not yet in production nothing is deployed there yet, the `usd-max-vault` branch is at Work-in-progress stage right now

danoctavian:
merging this, since no further comments

PR #130, opened 2025-05-08
update Provider with ynClisBnb strategy as an asset

yield-farmer-1:


danoctavian:
let's write a test for this as well in the provider test

danoctavian (approved):
LGTM

PR #136, opened 2025-05-12
bump BNB Max Vault to use Vault.VAULT_VERSION 0.3.0 and STRATEGY_VERS…

danoctavian:
…ION 0.2.0


script/config/WithdrawerConfig.sol
```diff
@@ -25,8 +25,11 @@ library WithdrawerConfig {
             uint8 decimals_ = 18;
             bool countNativeAsset_ = true;
             bool alwaysComputeTotalAssets_ = false;
+            uint256 defaultAssetIndex = 0;
 
-            vault.initialize(deployer, name, symbol, decimals_, countNativeAsset_, alwaysComputeTotalAssets_);
+            vault.initialize(
+                deployer, name, symbol, decimals_, countNativeAsset_, alwaysComputeTotalAssets_, defaultAssetIndex
```
yield-farmer-1, line 31: @danoctavian defaultAssetIndex is hardcoded and set directly to `0` in `eth-max-vault` PR. Semantically, it's same for both though

danoctavian: yes here as well, i just had a var

PR #148, opened 2025-06-11
Feature/usd max vault v1

yield-farmer-1:


test/mocks/MockProvider.sol
```diff
@@ -0,0 +1,32 @@
...
+
+/*
+    The Provider fetches state from other contracts.
+*/
+
+interface IBaseStrategy {
+    function STRATEGY_VERSION() external view returns (string memory);
+}
+
+// contract MockProvider is Provider {
```
danoctavian: let's delete this if not used

if we're not using it at all delete the whole file

test/mainnet/helpers/BaseTest.sol
```diff
@@ -0,0 +1,174 @@
...
+        vault.grantRole(vault.UNPAUSER_ROLE(), UNPAUSER);
+        vault.grantRole(vault.FEE_MANAGER_ROLE(), FEE_MANAGER);
+
+        // test cannot unpause vault without provider
+        vm.expectRevert();
+        vault.unpause();
+
+        vault.setProvider(address(provider));
+
+        vault.addAsset(address(wrappedUSDC), true);
```
danoctavian: this should be false

script/VerifyMaxVault.s.sol
```diff
@@ -27,21 +31,36 @@ contract VerifyMaxVault is BaseScript, Test {
...
+        assertEq(vaultProxy.baseWithdrawalFee(), 100000, "base withdrawal fee is invalid");
+        assertEq(vaultProxy.countNativeAsset(), false, "count native asset is invalid");
+        assertFalse(vaultProxy.alwaysComputeTotalAssets(), "always compute total assets is invalid");
+        IVault.AssetParams memory asset;
+        address[] memory assets = vaultProxy.getAssets();
+
+        assertEq(assets[0], MC.WRAPPED_USDC, "assets[0] is invalid");
+        asset = vaultProxy.getAsset(MC.WRAPPED_USDC);
+        assertEq(asset.decimals, 18, "asset[0].decimals is invalid");
+        assertEq(asset.active, true, "asset[0].active is invalid");
```
danoctavian: this should be false

danoctavian:
https://github.com/yieldnest/yieldnest-vault/blob/rwa-max-vault-release-candidate/test/mainnet/verifier.spec.sol#L32

  i suggest we add a test like this as well, this test keeps the Verify script valid at all times, i added it to ynRWAx

danoctavian:
can we force certain routes in paraswap?

right now it's not testing all code branches guaranteed as the nodejs API call may change routes

So there's multiple if branches in the ParaswapValidator that don't get exercised in a predictable way in the test

script/VerifyMaxVault.s.sol
yield-farmer-1: done

test/mainnet/helpers/BaseTest.sol
yield-farmer-1: done

test/mocks/MockProvider.sol
yield-farmer-1: deleted

danoctavian:
If routes are not forceable on paraswap let's have a unit test for ParaswapValidator.sol please

PR #161, opened 2025-07-11
Feat/performance fee

yield-farmer-1:


src/BaseVault.sol
```diff
@@ -775,11 +778,48 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
     }
 
     function _processAccounting() internal virtual {
-        uint256 totalBaseBalance = computeTotalAssets();
+        uint256 currentTotalBaseBalance = computeTotalAssets();
```
danoctavian: what if we separate this logic into a separate Contract that's a FeeModule that has mint power

And processAccounting just has a callback registered to call that's generic

and BaseVault calls into it.

so we separate that into a module that we can attach or not depending on if we have it on or not. 

so for example we won't have it there for the ynRWAx flex strategy and the clisBNB strategy that already takes fees on fund distribution

danoctavian: also keeps contract size under control

src/YnETHxFeeModule.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity ^0.8.24;
+
+import {OwnableUpgradeable, IERC4626, ERC20} from "src/Common.sol";
+import {IVault} from "src/interface/IVault.sol";
+import {YnETHx} from "src/YnETHx.sol";
+import {IFeeModule} from "src/interface/IFeeModule.sol";
+
+contract YnETHxFeeModule is OwnableUpgradeable, IFeeModule {
```
danoctavian: Shouldn't this have a generic names?

FeeModule
IVault VAULT:
uint256 VAULT_DECIMALS;


etc.

It can be used for any BaseVault

src/YnETHxFeeModule.sol
```diff
@@ -0,0 +1,67 @@
...
+
+    function initialize(address owner_, uint256 performanceFee_, address performanceFeeRecipient_)
+        external
+        initializer
+    {
+        __Ownable_init(owner_);
+        if (performanceFee_ > 1 ether) revert InvalidPerformanceFee();
+        performanceFee = performanceFee_;
+        performanceFeeRecipient = performanceFeeRecipient_;
+        maximumAccountedExchangeRate = IERC4626(address(YNETHX)).convertToAssets(1 ether);
```
danoctavian: this assumes a certain number of decimals since it converts 1 ether

src/YnETHxFeeModule.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity ^0.8.24;
+
+import {OwnableUpgradeable, IERC4626, ERC20} from "src/Common.sol";
+import {IVault} from "src/interface/IVault.sol";
+import {YnETHx} from "src/YnETHx.sol";
+import {IFeeModule} from "src/interface/IFeeModule.sol";
+
+contract YnETHxFeeModule is OwnableUpgradeable, IFeeModule {
+    // performance deniminated in ether(i.e. 1e18 = 100%)
```
danoctavian: deniminated -> denominated

src/YnETHxFeeModule.sol
```diff
@@ -0,0 +1,67 @@
...
+            uint256 exchangeRateAfterFee = IERC4626(address(YNETHX)).convertToAssets(1 ether);
+            if (exchangeRateAfterFee > maximumAccountedExchangeRate) {
+                maximumAccountedExchangeRate = exchangeRateAfterFee;
+            }
+        }
+    }
+
+    function setPerformanceFee(uint256 performanceFee_) external onlyOwner {
+        if (performanceFee_ > 1 ether) revert InvalidPerformanceFee();
+        performanceFee = performanceFee_;
```
danoctavian: emit events with prev and new

src/YnETHxFeeModule.sol
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    function setPerformanceFeeRecipient(address performanceFeeRecipient_) external onlyOwner {
+        performanceFeeRecipient = performanceFeeRecipient_;
```
danoctavian: even with prev and new

src/Vault.sol
```diff
@@ -107,6 +99,16 @@ contract Vault is BaseVault {
...
     }
 
+    function setFeeModule(address feeModule_) external virtual onlyRole(FEE_MANAGER_ROLE) {
+        _setFeeModule(feeModule_);
+    }
+
+    function _setFeeModule(address feeModule_) internal virtual {
+        FeeStorage storage feesStorage = _getFeeStorage();
+        feesStorage.feeModule = feeModule_;
+        emit SetFeeModule(feeModule_);
```
danoctavian: let's do prev and new, even if we haven't done it before

src/Vault.sol
```diff
@@ -107,6 +99,16 @@ contract Vault is BaseVault {
         _setBaseWithdrawalFee(baseWithdrawalFee_);
     }
 
+    function setFeeModule(address feeModule_) external virtual onlyRole(FEE_MANAGER_ROLE) {
```
danoctavian: is there any reason not to have these functions directly in BaseVault?

they are used there

src/BaseVault.sol
```diff
@@ -780,6 +784,25 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
...
         // solhint-disable-next-line not-rely-on-time
         emit ProcessAccounting(block.timestamp, totalBaseBalance);
+
+        address feeModule_ = feeModule();
+        if (feeModule_ != address(0)) {
+            IFeeModule(feeModule_).chargePerformanceFee();
+        }
+    }
+
+    function mintPerformanceFeeShares(address recipient, uint256 shares) external {
```
danoctavian: what if we just call this simply mint(,)

yield-farmer-1: mint function is already present in BaseVault and we can't use it because it takes deposit from caller. So, we need to declare different function with different name

src/Vault.sol
yield-farmer-1: updated this and also added new role to manage this function

src/Vault.sol
yield-farmer-1: done

src/YnETHxFeeModule.sol
yield-farmer-1: done

src/YnETHxFeeModule.sol
yield-farmer-1: done

src/YnETHxFeeModule.sol
yield-farmer-1: refactored it to use vault decimals

src/YnETHxFeeModule.sol
yield-farmer-1: done

src/YnETHxFeeModule.sol
yield-farmer-1: done

src/BaseVault.sol
danoctavian: right the erc4626, 

How about we call it `mintShares` simply

src/BaseVault.sol
```diff
@@ -780,6 +785,35 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
         _getVaultStorage().totalAssets = totalBaseBalance;
         // solhint-disable-next-line not-rely-on-time
         emit ProcessAccounting(block.timestamp, totalBaseBalance);
+
+        address feeModule_ = feeModule();
+        if (feeModule_ != address(0)) {
+            IFeeModule(feeModule_).chargePerformanceFee();
```
danoctavian: I suggest we make a simple naming change that just confines more generality.

We can use the IHooks name instead of IFeeModule

and this particular function would be hooks.afterProcessAccounting()

This follows the Uniswap V4 naming pattern.

https://github.com/Uniswap/v4-core/blob/main/src/PoolManager.sol#L178C36-L178C45

This way it follows a more general pattern and we can potentially have other functionality there without making the name unfitting.

So this involves mentioning fees explicitly here.

see the rename below as well.

src/BaseVault.sol
```diff
@@ -780,6 +785,35 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
...
+        _setFeeModule(feeModule_);
+    }
+
+    function _setFeeModule(address feeModule_) internal virtual {
+        FeeStorage storage feesStorage = _getFeeStorage();
+        emit SetFeeModule(feesStorage.feeModule, feeModule_);
+        feesStorage.feeModule = feeModule_;
+    }
+
+    function feeModule() public view returns (address) {
```
danoctavian: this should just return something of type IHooks if we do tha renaming as opposed to simply an address.

src/FeeModule.sol
```diff
@@ -0,0 +1,67 @@
...
+import {YnETHx} from "src/YnETHx.sol";
+import {IFeeModule} from "src/interface/IFeeModule.sol";
+
+contract FeeModule is OwnableUpgradeable, IFeeModule {
+    // performance denominated in ether(i.e. 1e18 = 100%)
+    uint256 public performanceFee;
+    address public performanceFeeRecipient;
+    IVault public immutable VAULT;
+    uint256 public immutable VAULT_DECIMALS;
+    uint256 public maximumAccountedExchangeRate;
```
danoctavian: Tis High-water-mark approach is uncommon in Defi.

and it will short-change us in case say there's a slashing event and then a lot more TVL comes in, no fees will be charged for the new growth of that new TVL.

i argue we shouldn't do it and just charge on harvest which is also easier to understand


Protocol | Vault Version | Performance Fee Charged On | High-Water Mark Support | Per-User Fee Basis? | Notes
-- | -- | -- | -- | -- | --
Yearn V2 | V2 | Yield realized during harvest | ❌ No | ❌ No | Global harvest-based; fees apply even during recovery
Yearn V3 | V3 | Strategy-level yield | ❌ No (by default) | ⚠️ Partial* | Vault can plug in custom fee logic, but no default HWM logic
Enzyme Finance | N/A | NAV increase over user’s entry value | ✅ Yes | ✅ Yes | Fully modular asset management with per-user HWM tracking
Sommelier | Structured | Yield vs deposit basis (some vaults) | ⚠️ Partial | ⚠️ Partial | Some strategies do HWM-like logic across deposit epochs
Ribbon Finance | Ribbon V1/V2 | Net vault gain per weekly epoch | ⚠️ Epoch HWM | ❌ No | Fee charged per round, no individual HWM tracked
Vesta Finance | Structured | Tranche-based performance | ⚠️ Epoch HWM | ❌ No | New epoch = new baseline; good for isolating post-slash gains
Beefy Finance | N/A | Harvested rewards | ❌ No | ❌ No | Fees taken at compound/harvest, not on net PPS gain
Convex Finance | N/A | Claimable CRV/CVX | ❌ No | ❌ No | No vault share price; fee on reward distribution only
Idle Finance | Best Yield | Yield earned | ❌ No | ❌ No | No HWM, fees taken from interest
Tokemak | Reactors | Yield streams | ⚠️ Epoch Tranches | ⚠️ Indirect | Epoch/tranche separation isolates fee base
Gearbox | Yield Strategies | Strategy PnL | ❌ No | ⚠️ Partial | Depends on integration, no built-in per-user HWM

src/FeeModule.sol
```diff
@@ -0,0 +1,67 @@
...
+
```
danoctavian: assuming we don't do HWM or using the all-time-high max,


consider what arugments you need to pass into the `afterProcessAccounting` call to make it really easy to calculate things here and simplify all computations

I think it won't need any extra storage

src/BaseVault.sol
```diff
@@ -775,11 +781,44 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
...
     function _processAccounting() internal virtual {
+        uint256 exchangeRateBefore = IERC4626(address(this)).convertToAssets(10 ** decimals());
+
         uint256 totalBaseBalance = computeTotalAssets();
 
         _getVaultStorage().totalAssets = totalBaseBalance;
         // solhint-disable-next-line not-rely-on-time
         emit ProcessAccounting(block.timestamp, totalBaseBalance);
+
+        address hooks_ = address(hooks());
```
danoctavian: nitpick, but the hooks() can be cast to an address just inside the if

so you can Read

```IHooks hooks_ = hooks();```

File: src/BaseVault.sol [hunk 15]
```diff
@@ -775,11 +781,44 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
...
+        if (hooks_ != address(0)) {
+            uint256 exchangeRateAfter = IERC4626(address(this)).convertToAssets(10 ** decimals());
+            uint256 totalSupplyBefore = totalSupply();
+            IHooks(hooks_).afterProcessAccounting(exchangeRateBefore, exchangeRateAfter, totalSupplyBefore);
```
danoctavian: i'd pass in totalBaseAssetsBefore and totalBaseAssetsAfter

instead of a transform which is convertToAssets, so that the hook gets raw data

and processes accordingly, instead of reversing the computation if it needs to get totalAssets for whatever reason

src/Hooks.sol
```diff
@@ -0,0 +1,61 @@
...
+        performanceFeeRecipient = performanceFeeRecipient_;
+    }
+
+    function afterProcessAccounting(uint256 exchangeRateBefore, uint256 exchangeRateAfter, uint256 totalSupplyBefore)
+        external
+    {
+        if (msg.sender != address(VAULT)) revert CallerNotVault();
+
+        if (exchangeRateAfter > exchangeRateBefore) {
+            uint256 yieldEarned = (exchangeRateAfter - exchangeRateBefore) * totalSupplyBefore / (10 ** VAULT_DECIMALS);
```
danoctavian: if we pass in raw totalBaseAssets

this computation can simply be a substraction

src/Hooks.sol
```diff
@@ -0,0 +1,61 @@
...
+        performanceFee = performanceFee_;
+        performanceFeeRecipient = performanceFeeRecipient_;
+    }
+
+    function afterProcessAccounting(uint256 exchangeRateBefore, uint256 exchangeRateAfter, uint256 totalSupplyBefore)
+        external
+    {
+        if (msg.sender != address(VAULT)) revert CallerNotVault();
+
+        if (exchangeRateAfter > exchangeRateBefore) {
```
danoctavian: also with raw totalBaseAssets this if would just compare that

src/Hooks.sol
```diff
@@ -0,0 +1,61 @@
...
+            uint256 yieldEarned = (exchangeRateAfter - exchangeRateBefore) * totalSupplyBefore / (10 ** VAULT_DECIMALS);
+            // dividing by 1 ether because performanceFee is denominated in ether(i.e. 1e18 = 100%)
+            uint256 feesAccrued = (yieldEarned * performanceFee) / 1 ether;
+
+            if (performanceFeeRecipient != address(0) && feesAccrued > 0) {
```
danoctavian: to reduce complexity here i'd not allow performanceFeeRecipient to be set to 0  at all and  assume that invariant.

Now there's 2 controls that toggle the fees on and off, setting `performanceFee_` to 0. or  setting `performanceFeeRecipient` to address(0). Having one mechanism is enough

So we should check `performanceFeeRecipient` is not zero when setting it and not have  it in the if clause here

src/YnETHx.sol
```diff
@@ -45,4 +45,8 @@ contract YnETHx is Vault {
         erc4626Storage._asset = IERC20(0x0000000000000000000000000000000000000000);
```
danoctavian: let's delete this file it's no longer used

src/Hooks.sol
```diff
@@ -0,0 +1,61 @@
...
+                uint256 sharesToMint = IERC4626(address(VAULT)).previewDeposit(feesAccrued);
```
danoctavian: so fee is calculated based on PPS after mint

this looks right

src/Hooks.sol
```diff
@@ -0,0 +1,61 @@
...
+        external
+        initializer
+    {
+        __Ownable_init(owner_);
+        if (performanceFee_ > 1 ether) revert InvalidPerformanceFee();
+        performanceFee = performanceFee_;
+        performanceFeeRecipient = performanceFeeRecipient_;
+    }
+
+    function afterProcessAccounting(uint256 exchangeRateBefore, uint256 exchangeRateAfter, uint256 totalSupplyBefore)
```
danoctavian: let's write unit tests for this specific function 

and also let's double down maing sure we test with:

the 6 decimals default asset  and the 18 decimal  shares vaults (example ynETHx)

the 6 decimals default asset and 18 decimals shares vault (example ynUSDx)

the 6 decimals default asset and 6 decimals shares vault (example Flex Strategy for ynRWAx)

src/Hooks.sol
```diff
@@ -0,0 +1,61 @@
...
+        if (exchangeRateAfter > exchangeRateBefore) {
+            uint256 yieldEarned = (exchangeRateAfter - exchangeRateBefore) * totalSupplyBefore / (10 ** VAULT_DECIMALS);
+            // dividing by 1 ether because performanceFee is denominated in ether(i.e. 1e18 = 100%)
+            uint256 feesAccrued = (yieldEarned * performanceFee) / 1 ether;
+
+            if (performanceFeeRecipient != address(0) && feesAccrued > 0) {
+                uint256 sharesToMint = IERC4626(address(VAULT)).previewDeposit(feesAccrued);
+                if (sharesToMint > 0) {
+                    VAULT.mintShares(performanceFeeRecipient, sharesToMint);
+                    emit PerformanceFeeCharged(performanceFeeRecipient, sharesToMint, feesAccrued);
```
danoctavian: might as well pack more info into the event, like totalAssetsBefore and totalAssetsAfter, and supply

src/BaseVault.sol
yield-farmer-1: done

src/BaseVault.sol
yield-farmer-1: done the changes

src/Hooks.sol
yield-farmer-1: done

src/Hooks.sol
yield-farmer-1: done

src/Hooks.sol
yield-farmer-1: done

src/YnETHx.sol
yield-farmer-1: removed

src/Hooks.sol
yield-farmer-1: added the tests

src/Hooks.sol
yield-farmer-1: added more details in event

src/BaseVault.sol
```diff
@@ -16,6 +16,11 @@ import {VaultLib} from "src/library/VaultLib.sol";
 
 import {IVault} from "src/interface/IVault.sol";
 import {IStrategy} from "src/interface/IStrategy.sol";
+import {FeeMath} from "src/module/FeeMath.sol";
+
+import {console} from "lib/forge-std/src/console.sol";
```
danoctavian: remove console

src/Hooks.sol
```diff
@@ -0,0 +1,60 @@
...
+        performanceFeeRecipient = performanceFeeRecipient_;
+    }
+
+    function afterProcessAccounting(uint256 totalAssetsBefore, uint256 totalAssetsAfter) external {
+        if (msg.sender != address(VAULT)) revert CallerNotVault();
+
+        if (totalAssetsAfter > totalAssetsBefore) {
+            uint256 yieldEarned = totalAssetsAfter - totalAssetsBefore;
+            // dividing by 1 ether because performanceFee is denominated in ether(i.e. 1e18 = 100%)
+            uint256 feesAccrued = (yieldEarned * performanceFee) / 1 ether;
```
danoctavian: let's use a clear constant called FEE_DENOMINATOR here = 1e18 declared publicly

src/BaseVault.sol
yield-farmer-1: removed

src/Hooks.sol
yield-farmer-1: done

danoctavian:
@yield-farmer-1  we also need to bump STRATEGY_VERSION since now it got changed by BaseVault changes implicitly

danoctavian:
a failure on test `testDonateOETHAndWithdraw` looks like error treshold got exceeded slightly, unrelated to PR

danoctavian:
This is improperly tested and verified please write proper tests for the formula

test/mainnet/invariants.spec.sol
```diff
@@ -262,9 +262,19 @@ contract VaultMainnetInvariantsTest is BaseIntegrationTest, TestHelper {
...
         {
             uint256 totalSupplyBefore = vault.totalSupply();
+            uint256 totalAssetsBefore = vault.totalAssets();
             vault.processAccounting();
             uint256 totalSupplyAfter = vault.totalSupply();
+            uint256 totalAssetsAfter = vault.totalAssets();
             performanceFeeSharesMinted = totalSupplyAfter - totalSupplyBefore;
+            if (performanceFeeSharesMinted > 0) {
+                assertApproxEqAbs(
+                    vault.convertToAssets(performanceFeeSharesMinted),
```
yield-farmer-1, line 294: asserts that performance fees shares minted is equivalent to the performance fees to be charged

danoctavian:
As discussed BaseStrategy does not implement Hooks properly 

and needs proper test coverage there for that as well.

src/interface/IVault.sol
```diff
@@ -34,10 +35,18 @@ interface IVault is IERC4626 {
...
+        /// @notice Whether the fee is overridden for the user
+        bool isOverridden;
+    }
+
     struct FeeStorage {
         /// @notice The base withdrawal fee in basis points (1e8 = 100%)
         uint64 baseWithdrawalFee;
-        mapping(address user => bool isExempted) withdrawalFeeExempted;
+        mapping(address user => OverriddenBaseWithdrawalFeeFields fields) overriddenBaseWithdrawalFee;
+        IHooks hooks;
```
danoctavian, line 52: why is this called feeStorage if it stores fee related things and hooks as well?

test/mainnet/basicfunctionality.spec.sol
```diff
@@ -400,8 +405,8 @@ contract VaultBasicFunctionalityTest is BaseIntegrationTest, TestHelper {
             "Withdrawer shares balance should remain unchanged"
         );
 
-        assertApproxEqAbs(
-            withdrawer.totalAssets(), withdrawerTotalBefore, 3, "Withdrawer total assets should remain unchanged"
+        assertApproxEqRel(
+            withdrawer.totalAssets(), withdrawerTotalBefore, 5e14, "Withdrawer total assets should remain unchanged"
```
danoctavian: let's not merge these kind of changes  please, increases deltas here from 3 wei to 0.005 is not justified

danoctavian: i'm creating a PR against eth-max-vault-release-candidate

danoctavian (approved):
lgtm

PR #171, opened 2025-07-29
Fix test set up

yield-farmer-1:


script/Contracts.sol
```diff
@@ -34,10 +34,10 @@ library MainnetContracts {
 
     address public constant WRAPPED_USDC = 0xdA7d2025c7f1f1A1d34AB3F4dF01102d0428E574;
 
-    address public constant YNUSDx = 0x3DB228FE836D99Ccb25Ec4dfdC80ED6d2CDdCB4b; // TODO: Update with deployed YnUSDx
+    address public constant YNUSDx = 0x3DB228FE836D99Ccb25Ec4dfdC80ED6d2CDdCB4b;
 
-    address public constant PROVIDER = address(1234567890); // Mock value
```
danoctavian, line 39: i think we should actually delete PROVIDER and BUFFER and always use the dynamic value to avoid confusions here

yield-farmer-1: done the changes

danoctavian (approved):
lgtm

PR #175, opened 2025-08-05
Feat/withdrawal fee to recipient

yield-farmer-1:
* The tests assert that shares minted to fee recipient is equivalent to withdrawal fee that needs to be charged on withdraw and redeem.

### DIFF
--- a/src/BaseVault.sol
+++ b/src/BaseVault.sol
@@ -315,13 +315,13 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
      * @param assets The amount of assets to withdraw.
      * @param receiver The address of the receiver.
      * @param owner The address of the owner.
-     * @return sharesWithFee The equivalent amount of shares with fee.
+     * @return shares The equivalent amount of shares.
      */
     function withdraw(uint256 assets, address receiver, address owner)
         public
         virtual
         nonReentrant
-        returns (uint256 sharesWithFee)
+        returns (uint256 shares)
     {
         if (paused()) {
             revert Paused();
@@ -330,14 +330,14 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
         if (assets > maxAssets) {
             revert ExceededMaxWithdraw(owner, assets, maxAssets);
         }
-        sharesWithFee = previewWithdraw(assets);
+        shares = previewWithdraw(assets);
 
         IHooks hooks_ = hooks();
         if (address(hooks_) != address(0)) {
             hooks_.beforeWithdraw(assets, _msgSender());
         }
 
-        _withdraw(_msgSender(), receiver, owner, assets, sharesWithFee);
+        _withdraw(_msgSender(), receiver, owner, assets, shares);
     }
 
     /**
--- a/src/BaseVault.sol
+++ b/src/BaseVault.sol
@@ -331,15 +331,13 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
             revert ExceededMaxWithdraw(owner, assets, maxAssets);
         }
         sharesWithFee = previewWithdraw(assets);
-        uint256 sharesWithoutFee = _previewWithdrawWithoutFee(assets);
-
-        _withdraw(_msgSender(), receiver, owner, assets, sharesWithFee);
 
         IHooks hooks_ = hooks();
         if (address(hooks_) != address(0)) {
-            uint256 sharesToMint = sharesWithFee - sharesWithoutFee;
-            hooks_.afterWithdraw(sharesToMint);
+            hooks_.beforeWithdraw(assets, _msgSender());
         }
+
+        _withdraw(_msgSender(), receiver, owner, assets, sharesWithFee);
     }
 
     /**
@@ -367,8 +365,7 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
 
         IHooks hooks_ = hooks();
         if (address(hooks_) != address(0)) {
-            uint256 sharesToMint = _feeOnTotal(shares);
-            hooks_.afterRedeem(sharesToMint);
+            hooks_.afterRedeem(shares, _msgSender());
         }
     }
 
--- a/src/BaseVault.sol
+++ b/src/BaseVault.sol
@@ -315,13 +315,13 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
      * @param assets The amount of assets to withdraw.
      * @param receiver The address of the receiver.
      * @param owner The address of the owner.
-     * @return shares The equivalent amount of shares.
+     * @return sharesWithFee The equivalent amount of shares with fee.
      */
     function withdraw(uint256 assets, address receiver, address owner)
         public
         virtual
         nonReentrant
-        returns (uint256 shares)
+        returns (uint256 sharesWithFee)
     {
         if (paused()) {
             revert Paused();
@@ -330,9 +330,16 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
         if (assets > maxAssets) {
             revert ExceededMaxWithdraw(owner, assets, maxAssets);
         }
-        shares = previewWithdraw(assets);
+        sharesWithFee = previewWithdraw(assets);
+        uint256 sharesWithoutFee = _previewWithdrawWithoutFee(assets);
 
-        _withdraw(_msgSender(), receiver, owner, assets, shares);
+        _withdraw(_msgSender(), receiver, owner, assets, sharesWithFee);
+
+        IHooks hooks_ = hooks();
+        if (address(hooks_) != address(0)) {
+            uint256 sharesToMint = sharesWithFee - sharesWithoutFee;
+            hooks_.afterWithdraw(sharesToMint);
+        }
     }
 
     /**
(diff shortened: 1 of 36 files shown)


src/BaseVault.sol
```diff
@@ -330,9 +330,16 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
...
+        sharesWithFee = previewWithdraw(assets);
+        uint256 sharesWithoutFee = _previewWithdrawWithoutFee(assets);
 
-        _withdraw(_msgSender(), receiver, owner, assets, shares);
+        _withdraw(_msgSender(), receiver, owner, assets, sharesWithFee);
+
+        IHooks hooks_ = hooks();
+        if (address(hooks_) != address(0)) {
+            uint256 sharesToMint = sharesWithFee - sharesWithoutFee;
+            hooks_.afterWithdraw(sharesToMint);
```
danoctavian: when adding these hooks in please think more generally as opposed to just thinking how to solve the current immediate problem of redirecting the withdrawal fee.

```uint256 sharesWithoutFee = _previewWithdrawWithoutFee(assets);```

ideally this line doesn't exist here as its specific to fees.

Let's think please if there's a more general solution here where the hook. receives enough info to infer the necessary information to compute the shares to be minted

Right now the fee related logic is baked into the BaseVault which defeats the purpose of making this separated Hooks module  as we put something specific to fees in the top BaseVault regardless

Also think how this hook can be used for other purposes by taking inspiration from the Uniswap hooks.

What if you had other problems you wanted to solve with the hooks?

would passing in this very specific `sharesToMint` parameter be enough?

File: src/BaseVault.sol [hunk 2]
```diff
@@ -332,6 +332,11 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
         }
         shares = previewWithdraw(assets);
 
+        IHooks hooks_ = hooks();
+        if (address(hooks_) != address(0)) {
+            hooks_.beforeWithdraw(assets, _msgSender());
```
danoctavian, line 337: this is not _msgSender() should be owner

src/Hooks.sol
```diff
@@ -59,12 +89,51 @@ contract Hooks is OwnableUpgradeable, IHooks {
...
     }
 
+    /**
+     * @notice Callback from the Vault before withdraw
+     * @param assets The amount of assets to be withdrawn
+     * @param user The address of the user withdrawing
+     */
+    function beforeWithdraw(uint256 assets, address user) external onlyVault {
+        Vault vault_ = Vault(payable(address(VAULT)));
+        if (vault_.withdrawalFeeExempted(user)) return;
```
danoctavian: this check is not needed and it should not be used as it assumes it's vault, and it won't work with BaseStrategy

if withdrawalFeeExempted then fee is 0

src/Hooks.sol
```diff
@@ -59,12 +89,51 @@ contract Hooks is OwnableUpgradeable, IHooks {
...
+    }
+
+    /**
+     * @notice Callback from the Vault after redeem
+     * @param shares The amount of shares to be redeemed
+     * @param user The address of the user redeeming
+     */
+    function afterRedeem(uint256 shares, address user) external onlyVault {
+        Vault vault_ = Vault(payable(address(VAULT)));
+        if (vault_.withdrawalFeeExempted(user)) return;
```
danoctavian: same here this should not be used

src/Hooks.sol
danoctavian: as pointed out by @yield-farmer-1  i take that back:

the msg.sender checkinside of  feeOnTotal and feeOnRaw means you msg.sender is Hooks and it misapplies the fee override.

src/BaseVault.sol
```diff
@@ -357,6 +362,11 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
         }
         assets = previewRedeem(shares);
         _withdraw(_msgSender(), receiver, owner, assets, shares);
+
+        IHooks hooks_ = hooks();
+        if (address(hooks_) != address(0)) {
+            hooks_.afterRedeem(shares, _msgSender());
```
danoctavian: afterRedeem calls _feeOnTotal which takes as a parameter a quantity in assets but this passes in shares

If by some math trick is correct, it needs a very well explained expanded explanation, as it's not obvious  why it is correct

src/Hooks.sol
```diff
@@ -59,12 +89,51 @@ contract Hooks is OwnableUpgradeable, IHooks {
...
+        uint256 sharesToMint = vault_._feeOnTotal(shares);
```
danoctavian: _feeOnTotal takes as a parameter assets, not obvious why this ok here needs clarity, as shares are passed in

src/Hooks.sol
danoctavian: still we need to cover this leak in the abstraction as this is not idealm since it repeats logic contained in fee calculation

src/BaseVault.sol
danoctavian: since we decided to implement hooks, we should go all the way here and provide full support.

In the case of beforeWithdraw you can pass in:

(caller, receiver, owner, assets, shares) - for the full context to be available. to the hook.

danoctavian: Analogously for for afterRedeem

(caller, receiver, owner, assets, shares)

danoctavian: also instead of an assymetrical implementation lets do before and after equivalents in the case of both.

And deposit hooks as well.

The deposit path may not benefit from a deposit/mint distinction, as there's no fee logic or any path dependant thing, but do argue otherwise if you see anything useful.

In the case of a deposit hook, seems like the whole context is

```        address asset_,
        address caller,
        address receiver,
        uint256 assets,
        uint256 shares,
        uint256 baseAssets```

File: src/BaseVault.sol [hunk 7]
```diff
@@ -332,6 +332,11 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
         }
         shares = previewWithdraw(assets);
 
+        IHooks hooks_ = hooks();
+        if (address(hooks_) != address(0)) {
```
danoctavian: i'm wondering if for cleanliness we mandate a hooks implementation so if there isn't one set it reverts. and there's a NOOP hook to attach.

And there's no check for adress 0.

the address(0) check is a toggle that helps i guess to reduce gas usage.

I do notice uni v4 doesn't check for address Zero, and they usually care about gas usage more than YieldNest as a DEX with very frequent swaps. beyond cleaner code and less assumptions, i do wonder what else was the reason there

src/BaseVault.sol
```diff
@@ -19,6 +19,7 @@ import {IStrategy} from "src/interface/IStrategy.sol";
 import {FeeMath} from "src/module/FeeMath.sol";
 
 import {IHooks} from "src/interface/IHooks.sol";
+import {HooksLib} from "src/module/HooksLib.sol";
```
danoctavian: should be in library directory rather next to VaultLib

src/Hooks.sol
```diff
@@ -5,7 +5,13 @@ import {OwnableUpgradeable, IERC4626, ERC20} from "src/Common.sol";
 import {IVault} from "src/interface/IVault.sol";
```
danoctavian: i'd argue Hooks fits better in the module directory

src/module/HooksLib.sol
```diff
@@ -0,0 +1,269 @@
...
+        AFTER_WITHDRAW,
+        BEFORE_PROCESS_ACCOUNTING,
+        AFTER_PROCESS_ACCOUNTING
+    }
+
+    // @notice Checks if the hook has the permission set for the given flag to be called
+    // @param self The hooks contract
+    // @param flag The flag to check
+    // @return True if the hook has the permission, false otherwise
+    function hasPermission(IHooks self, Flags flag) internal view returns (bool) {
```
danoctavian: the naming here could be better something like 

function hookEnabled(IHooks self, Flags flag)

and the enum is better called HookType

Flag and Permission are too generic and not related to the context.

src/interface/IHooks.sol
```diff
@@ -4,6 +4,21 @@ pragma solidity ^0.8.24;
 import {IVault} from "src/interface/IVault.sol";
 
 interface IHooks {
+    // @notice Permissions struct for the hooks
+    // @dev Each flag is a boolean value that indicates if the corresponding function of hook has the permission to be called from vault
+    struct Permissions {
```
danoctavian: is this actually better than a mapping in storage?

Is this more gas efficient or better in any way vs having a mapping from enum to boolean?

what are the arguments to use a struct with a bool for each thing?

src/Hooks.sol
```diff
@@ -15,59 +21,290 @@ contract Hooks is OwnableUpgradeable, IHooks {
...
+        /**
+         * totalBaseBalanceAfterAccounting*
+         */
+        uint256
+    )
+        /**
+         * totalBaseBalanceBeforeAccounting*
+         */
+        external
+        onlyVault
```
danoctavian: i propose we decouple the caller restriction from the vault

so we have onlyOwner

so that we can compose hooks in sequence with a Hook composer down the line.

so for example i can add a hook that adds some invariant checks and it stacks with this hook

src/interface/IHooks.sol
danoctavian: i'm guessing it's because all bools fit into one slot and then the if/else loop is quite cheap

yield-farmer-1: yes. that's the reason for it

src/BaseVault.sol
yield-farmer-1: done

src/Hooks.sol
yield-farmer-1: done

src/module/HooksLib.sol
yield-farmer-1: renamed

src/interface/IHooks.sol
```diff
@@ -18,13 +33,44 @@ interface IHooks {
     );
     event SetPerformanceFee(uint256 oldFee, uint256 newFee);
```
danoctavian: this interface should not have anything about fees, fee functions fee events

That's specific to the Fee Hook

This interface should only be a IHook Interface

Can define a separate interface for those things and have the concrete implementation use both

yield-farmer-1: fixed

PR #177, opened 2025-08-05
deploy withdrawer + fxsave strategy

danoctavian:


danoctavian:
Screenshot 2025-08-08 at 10 49 40 PM


@yield-farmer-1  if we add before and after hooks for symmetry, i believe there may be a more elegant solution that uses transient storage to calculate how many shares to mint, and minting them in the after call.

May not be needed, but if it makes it easier you could use tstore to do that.

PR #181, opened 2025-08-20
Fix async withdrawals tests

danoctavian:


test/mainnet/BaseWithdrawerTest.sol
```diff
@@ -146,9 +147,21 @@ abstract contract BaseWithdrawerMainnetTest is BaseIntegrationTest, TestHelper {
         vm.assume(amount > 1e6);
```
yield-farmer-1, line 147: maybe it's better that we should use `bound` instead of `vm.assume` when refactoring old tests. Because assume will just skip lot of tests if inputs are out of bound

yield-farmer-1 (approved):

PR #182, opened 2025-08-23
add withdrawAsset to BaseVault

danoctavian:
BaseVault.withdrawAsset is a permissioned function that is used as a building block for processing queued withdrawal requests for Max vaults.

The goal there would be to be able to redeem arbitrary assets from the vault, to funnel them to withdrawers.

The role is to be assigned to a peripheral YieldNest protocol contract.

the benefit here is the following:

1.you can fulfill large redemptions faster eg. you don't need to redeem wstETH if the redeemer is ok with it.
 redeeming wstETH to service withdrawals in the max vault (Withdrawer.sol) is yield-inneficient as it affects all stakers since they all take the yield hit.

2.you protect the ynETHx yield becase wstETH in withdrawal state is not earning yield
3.you can build queue logic on top of that as a separate peripheral system

### DIFF
--- a/src/BaseVault.sol
+++ b/src/BaseVault.sol
@@ -404,6 +404,16 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
     }
 
     /**
+     * @notice Returns whether a given asset exists in the asset list of the vault.
+     * @param asset_ The address of the asset.
+     */
+    function hasAsset(address asset_) public view virtual returns (bool) {
+        AssetStorage storage assetStorage = _getAssetStorage();
+        AssetParams memory assetParams = assetStorage.assets[asset_];
+        return assetStorage.list[assetParams.index] == asset_;
+    }
+
+    /**
      * @notice Returns the function rule for a given contract address and function signature.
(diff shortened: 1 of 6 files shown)


danoctavian:
build failures unrelated to to code deployed, it's related to rule change in production, will address

src/BaseVault.sol
```diff
@@ -521,6 +531,58 @@ abstract contract BaseVault is IVault, ERC20PermitUpgradeable, AccessControlUpgr
...
+     */
+    function _withdrawAsset(
+        address asset_,
+        address caller,
+        address receiver,
+        address owner,
+        uint256 assets,
+        uint256 shares
+    ) internal virtual {
+        if (!hasAsset(asset_)) revert InvalidAsset(asset_);
```
danoctavian, line 603: there's actually a bug related to this check in the current BaseStrategy where you can delete an asset and still withdraw it.

danoctavian: Fixed here: https://github.com/yieldnest/yieldnest-vault/pull/185

yield-farmer-1 (approved):

PR #183, opened 2025-08-24
fix test failures due to config changes

danoctavian:


test/mainnet/basicfunctionality.spec.sol
```diff
@@ -62,9 +62,6 @@ contract VaultBasicFunctionalityTest is BaseIntegrationTest, TestHelper {
         // verify withdrawer config
         VaultVerification.verifyWithdrawerConfiguration(vault, withdrawer);
 
-        // verify withdrawer roles
-        VaultVerification.verifyWithdrawerRules(withdrawer);
```
danoctavian, line 66: remove since these always change

script/VerifyMaxVault.s.sol
```diff
@@ -102,9 +102,6 @@ contract VerifyMaxVault is BaseScript, Test {
         // Verify vault configuration using VaultVerification library
         VaultVerification.verifyVaultConfiguration(vault, withdrawer);
 
-        // Verify processor rules
-        VaultVerification.verifyRules(vault);
```
danoctavian, line 106: rules are too dynamic to assert in tests

test/mainnet/invariants.spec.sol
```diff
@@ -1192,7 +1192,7 @@ contract VaultMainnetInvariantsTest is BaseIntegrationTest, TestHelper {
         }
 
         {
-            _processWithdrawAsset(address(withdrawer), MC.WETH, amount);
```
danoctavian, line 1195: no need for depositAsset for WETH, it works with deposit(..)

